PR #12, opened 2019-10-10
Succinct error contexts

codygibb:
Recommend keeping error contexts succinct.

### DIFF
--- a/style.md
+++ b/style.md
@@ -729,8 +729,53 @@ There are three main options for propagating errors if a call fails:
   specific error case.
 
 It is recommended to add context where possible so that instead of a vague
-error such as "connection refused", you get more useful errors such as "failed to
-call service foo: connection refused".
+error such as "connection refused", you get more useful errors such as
+"call service foo: connection refused".
+
+When adding context to returned errors, keep the context succinct by avoiding
+phrases like "failed to", which state the obvious and pile up as the error
+percolates up through the stack:
+
+<table>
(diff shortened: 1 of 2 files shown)


codygibb:
SGTM. I'll expand on this a bit more.

style.md
```diff
@@ -729,8 +729,51 @@ There are three main options for propagating errors if a call fails:
...
-error such as "connection refused", you get more useful errors such as "failed to
-call service foo: connection refused".
+error such as "connection refused", you get more useful errors such as
+"call service foo: connection refused".
+
+When adding context to returned errors, keep the context succinct by avoiding
+phrases like "failed to", which state the obvious and pile up as the error
+percolates up through the stack:
+
+<table>
```
prashantv: this example is resulting in scrollbars for the code snippets:
https://github.com/uber-go/guide/blob/4f628ff1b1130b013bb2ab46cf0b585fa6cb2d5a/style.md#error-wrapping

can you play with the example to still make the point that "failed to" gets repetitive, but also reducing the line length so it fits in the default github width?

codygibb: I was irked by it as well... Yeah I'll get this looking nice.

codygibb: I placed a somewhat questionable line break in the code. Without it we have a tiny bit of scroll bar. Lmk if you'd prefer a little scroll over the line break.

prashantv: lgtm, I think I'd prefer to avoid the scrollbar.

PR #16, opened 2019-10-11
Add benchmarks for strconv vs fmt guideline

prashantv:


style.md
```diff
@@ -931,15 +931,30 @@ When converting primitives to/from strings, `strconv` is faster than
 <tr><td>
 
 ```go
-var i int = ...
-s := fmt.Sprint(i)
+for i := 0; i < b.N; i++ {
+  s := fmt.Sprint(r.Int())
```
abhinav: Should this be `rand.Int()`, or perhaps include initializing `r := rand.New(..)` to make this more 'runnable'?

prashantv: I used `r := rand.New(..)` since `rand.Int()` has some overhead with locks, but I think keeping this example simpler is more important, even at the expense of some additional overhead from rand. Updated to `rand.Int()` and updated benchmark numbers which are a few nanoseconds slower.

PR #20, opened 2019-10-11
Reduce Scope: Fix FD leak

### LINKED ISSUE #19: "Reduce Scope of Variables" second good example can leak an open file
Copying my comment from HN thread:

I find it amusing that the advice in the style guide gives a good example contradicting another good example, and contains a subtle bug.

In the "Reduce Scope of Variables", second good example leaks an open file when WriteString fails, because it doesn't follow the own advice of "Defer to Clean Up" if you are curious.

(Handling that properly with a defer is a bit more tricky - something like https://play.golang.org/p/l1PeWM3Tisg).

abhinav:
As pointed out in #19, the second Good example for "Reduce Scope of
Variables" leaks an open file descriptor if `io.WriteString` fails.

Further, both sets of examples are manually closing files instead of
using `defer` as pointed out elsewhere in the guide.

As these examples just need a pair of functions, one returning just an
error, another returning a value and an error, I've switched them to
using `ioutil.ReadFile`, `ioutil.WriteFile`, and a plausible
`Config.Decode([]byte) error` method on a `cfg` variable assumed to be
in-scope.

Resolves #19

### DIFF
--- a/style.md
+++ b/style.md
@@ -1762,7 +1762,7 @@ conflicts with [Reduce Nesting](#reduce-nesting).
 <tr><td>
 
 ```go
-err := f.Close()
+err := ioutil.WriteFile(name, data, 0644)
 if err != nil {
  return err
 }
@@ -1771,7 +1771,7 @@ if err != nil {
 </td><td>
 
 ```go
-if err := f.Close(); err != nil {
+if err := ioutil.WriteFile(name, data, 0644); err != nil {
  return err
 }
 ```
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -1802,16 +1802,16 @@ if f, err := os.Open("f"); err == nil {
...
    return err
 }
 
-if _, err := io.WriteString(f, "data"); err != nil {
+if err := cfg.Decode(data); err != nil {
   return err
 }
 
-return f.Close()
+return nil
```
prashantv: maybe we have a comment like
```
// use cfg for some operations

return nil
```

otherwise, this would probalby just be `return cfg.Decode(data)`

abhinav: 👍 going to print it as a placeholder for "Do stuff with cfg"

PR #21, opened 2019-10-12
Don't embed mutex in exported struct

betamos:
As per section #zero-value-mutexes-are-valid: "For exported types, use a private lock."

### DIFF
--- a/style.md
+++ b/style.md
@@ -361,15 +361,15 @@ state.
 
 ```go
 type Stats struct {
-  sync.Mutex
+  mu sync.Mutex
 
   counters map[string]int
 }
 
 // Snapshot returns the current stats.
 func (s *Stats) Snapshot() map[string]int {
-  s.Lock()
-  defer s.Unlock()
+  s.mu.Lock()
+  defer s.mu.Unlock()
 
   return s.counters
 }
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -382,14 +382,14 @@ snapshot := stats.Snapshot()
 
 ```go
 type Stats struct {
-  sync.Mutex
+  mu sync.Mutex
```
mway: nit: let's remove this line now that the mutex is no longer embedded

abhinav:
```suggestion
```

style.md
```diff
@@ -361,15 +361,15 @@ state.
...
```
abhinav:
```suggestion
```

abhinav:
@betamos Would you mind signing the CLA? Thanks!

betamos:
I'm awaiting approval for signing the CLA from the company I work at (sigh). Hopefully resolved in the early week.

betamos:
This was a hurdle :(. Essentially me signing the CLA requires a high level approval that I cannot easily get. Perhaps you can accept it the PR without (or I would not mind if you discard it and fix it yourselves instead)? Essentially, a small bug fix or a change like this might not be intellectual property at all, hence a CLA would not be required on your end. But I'm not a lawyer.

prashantv:
No worries, thanks for the contribution @betamos. I'll take over this in #45.

PR #27, opened 2019-10-13
Mutexes should only be embedded in unexported

### LINKED ISSUE #23: Clarify embed guidance to address exported type APIs
Expanding on #21, we should make a distinction for exported and unexported types, and call out advice to avoid bleeding implementation details through APIs when embedding.

cc @abhinav @prashantv

prashantv:
More explicitly call out that mutexes should only be embedded
in unexported types.

Related to #23.

### DIFF
--- a/style.md
+++ b/style.md
@@ -296,7 +296,7 @@ func (m *SMap) Get(k string) string {
 </tr>
 <tr>
 <td>Embed for private types or types that need to implement the Mutex interface.</td>
-<td>For exported types, use a private lock.</td>
+<td>For exported types, use a private field.</td>
 </tr>
 
 </tbody></table>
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -238,16 +238,18 @@ mu.Lock()
 </td></tr>
 </tbody></table>
 
-If you use a struct by pointer, then the mutex can be a non-pointer field or,
-preferably, embedded directly into the struct.
+If you use a struct by pointer, then the mutex can be a non-pointer field.
+
+Unexported structs that use a mutex to protect fields of the struct should embed
```
abhinav: s/should/may since you may not want to make the methods part of the unexported struct's 'public' interface

```suggestion
Unexported structs that use a mutex to protect fields of the struct may embed
```

PR #42, opened 2019-10-14
add case

StarpTech:


mway:
Hey @StarpTech, thanks for the suggestion! To date, we haven't observed this enough to feel that it's a common issue worth calling out with guidance, so we're not going to include this right now. If it becomes a more frequent issue, or if we start to notice subtle or insidious bugs as a result, we may re-evaluate.

Thanks again, and please continue to make other suggestions!

PR #43, opened 2019-10-14
add style

StarpTech:


abhinav:
Hey, this advice conflicts with [Avoid Mutable Globals][1].

[1]: https://github.com/uber-go/guide/blob/master/style.md#avoid-mutable-globals

We agree that `init()` should be avoided (#72) but exported
initialization functions are problematic too. We recommend avoiding
global state when possible. There are scenarios where `init()` is valid
(some mentioned in #72) and we would not advise using an exported
initialization function for those.

StarpTech:
> We recommend avoiding global state when possible

I agree but `init` is part of the language and therefore worth noting.

> There are scenarios where init() is valid
(some mentioned in #72) and we would not advise using an exported
initialization function for those.

I refer to them in the examples.

abhinav:
Sorry, let me clarify.

`init()` is sometimes used to maintain global state. We advise against
that.

However, when `init()` is used in safe ways to initialize a module,
like to hold deterministic pre-computed information, we recommend using
`init()` itself, not an exported package initialization function.

PR #50, opened 2019-10-18
update a comment for snapshot example

TangoEnSkai:
I know this is very small thing, but I just wondering this comment is clearer and easier to understand.


abhinav:
Hey, it looks like there are a bunch of unrelated changes in this commit. You're only trying to change the comment, right? The diff shows a bunch of other things including the TOC. Can you remove those?

TangoEnSkai:
thanks. oh yes, I realise my linter make the links in that tree-structure. let me revise it. sorry.

style.md
```diff
@@ -375,8 +375,8 @@ func (s *Stats) Snapshot() map[string]int {
   return s.counters
 }
 
-// snapshot is no longer protected by the mutex, so any
-// access to the snapshot is racy.
+// snapshot is no longer protected by the mutex,
+// so any access to the snapshot is subject to race condition.
```
mway: spelling: "condition" should be "conditions"

abhinav:
```suggestion
// so any access to the snapshot is subject to race conditions.
```

TangoEnSkai:
thanks for your time!

style.md
prashantv: - I think there's no need to move `so any`
 - optionally can call out data races specifically

```
// snapshot is no longer protected by the mutex, so any
// access to the snapshot is subject to data races.
```

TangoEnSkai: I like it. I will update it very soon. thanks

TangoEnSkai:
@prashantv thanks for your comment. just updated as your recommendation.

PR #51, opened 2019-10-18
Wrapping errors with %w

topac:
as stated here https://blog.golang.org/go1.13-errors

### DIFF
--- a/style.md
+++ b/style.md
@@ -739,7 +739,7 @@ percolates up through the stack:
 s, err := store.New()
 if err != nil {
     return fmt.Errorf(
-        "failed to create new store: %s", err)
+        "failed to create new store: %w", err)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mway:
Hey, @topac!

We won't be able too make this a universal guideline (at least for a while) due to `go vet` failures in pre-1.13 releases of Go:

```
$ go vet ./
tmp.go:10:8: Errorf format %w has unknown verb w
```

We could add a note that this guidance applies to go1.13+, but we may want to take a little more time to let error unwrapping bake so that we don't prematurely introduce guidance.

@abhinav thoughts?

abhinav:
Agreed that it may be too early to make this recommendation.

Also, I don't think we want to add `%w` as a universal guidance. The error
wrapping guidance will have to change to suggest `%w` when exposing the
underlying error is desirable, versus `%v` when it is not. This is mentioned
in the [Whether to Wrap section](https://blog.golang.org/go1.13-errors#TOC_3.4.) of the blog post as well.

topac:
@abhinav you're right, you may want not to expose the underlying error. Closing this for now.

PR #54, opened 2019-10-20
Expand on embedding

### LINKED ISSUE #23: Clarify embed guidance to address exported type APIs

mway:
Fixes #23.

### DIFF
--- a/style.md
+++ b/style.md
@@ -2185,6 +2185,132 @@ type Client struct {
 </td></tr>
 </tbody></table>
 
+Embedding should provide tangible benefit, like adding or augmenting
+functionality in a semantically-appropriate way. It should do this with zero
+adverse user-facing effects (see also: [Avoid Embedding Types in Public Structs]).
+
+  [Avoid Embedding Types in Public Structs]: #avoid-embedding-types-in-public-structs
+
+Embedding **should not**:
+
+- Be purely cosmetic or convenience-oriented.
+- Make outer types more difficult to construct or use.
+- Affect outer types' zero values. If the outer type has a useful zero value, it
+  should still have a useful zero value after embedding the inner type.
+- Expose unrelated functions or fields from the outer type as a side-effect of
+  embedding the inner type.
+- Expose unexported types.
+- Affect outer types' copy semantics.
+- Change the outer type's API or type semantics.
+- Embed a non-canonical form of the inner type.
+- Expose implementation details of the outer type.
+- Allow users to observe or control type internals.
+- Change the general behavior of inner functions through wrapping in a way that
+  would reasonably surprise users.
+
+Simply put, embed consciously and intentionally. A good litmus test is, "would
+all of these exported inner methods/fields be added directly to the outer type";
+if the answer is "some" or "no", don't embed the inner type - use a field
+instead.
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+type A struct {
+    // Bad: A.Lock() and A.Unlock() are
+    //      now available, provide no
+    //      functional benefit, and allow
+    //      users to control details about
+    //      the internals of A.
+    sync.Mutex
+}
+```
+
+</td><td>
+
+```go
+type countingWriteCloser struct {
+    // Good: Write() is provided at this
+    //       outer layer for a specific
+    //       purpose, and delegates work
+    //       to the inner type's Write().
+    io.WriteCloser
+
+    count int
+}
+
+func (w *countingWriteCloser) Write(bs []byte) (int, error) {
+    w.count += len(bs)
+    return w.WriteCloser.Write(bs)
+}
+```
+
+</td></tr>
+<tr><td>
+
+```go
+type Book struct {
+    // Bad: pointer changes zero value usefulness
+    io.ReadWriter
+
+    // other fields
+}
+
+// later
+
+var b Book
+b.Read(...)  // panic: nil pointer
+b.String()   // panic: nil pointer
+b.Write(...) // panic: nil pointer
+```
+
+</td><td>
+
+```go
+type Book struct {
+    // Good: has useful zero value
+    bytes.Buffer
+
+    // other fields
+}
+
+// later
+
+var b Book
+b.Read(...)  // ok
+b.String()   // ok
+b.Write(...) // ok
+```
+
+</td></tr>
+<tr><td>
+
+```go
+type Client struct {
+    sync.Mutex
+    sync.WaitGroup
+    bytes.Buffer
+    url.URL
+}
+```
+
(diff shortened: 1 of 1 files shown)


mway:
if anyone wants to suggest some better examples, feel free.

prashantv:
Thanks for putting this together -- can we separate out the recommendations for be-consistent vs embedding?

mway:
> Thanks for putting this together -- can we separate out the recommendations for be-consistent vs embedding?

Of course - done in #58.

style.md
```diff
@@ -1557,7 +1557,237 @@ type Client struct {
...
 </tbody></table>
 
-### Use Field Names to initialize Structs
+As a general rule of thumb, embedding should not make types more difficult to
+use, and should be viewed as an "is-a" declaration: the embedding (outer) type
+can function as the embedded (inner) type, and therefore "is a" form of the
+inner type. A type embedding a `sync.Mutex` declares to users that it "is a"
+mutex; a type embedding a `bytes.Buffer` declares that it "is a" buffer.
+
+Embedded types' methods or fields should never be shadowed by first-class
```
prashantv: Never seems like a strong recommendation, I've often found that embedding + shadowing can be useful as a "decorator".

E.g., let's say you have a `io.WriteCloser`, you might add a decorator that tracks how many bytes `Write` was called on:
```go
type countingWriteCloser struct {
  io.WriteCloser
  
  count int
}

func (w *countingWriteCloser) Write(bs []byte) (int, error) {
  w.count += len(bs)
  return w.WriteCloser.Write(bs)
}
```

In tihs case, it'd be reasonable not to embed as well, but want to get your thoughts on this sort of use

style.md
```diff
@@ -1557,7 +1557,237 @@ type Client struct {
...
+methods or fields defined as part of the embedding type, and embedding should
+never be used for callsite brevity or other cosmetic convenience.
+
+When embedding, visibility and type semantics should be taken into account:
+because exported fields and methods of an embedded type will also be exported
+by the embedding type, careless use can bleed implementation details through
+types' API boundaries, expose internal type faculties that can be used in an
+unsafe or unknown way, or change the semantics of the embedding type altogether.
+
+All of the guidance in this section applies primarily to embedding types that
```
prashantv: Should we say an exported type should not embed an unexported type?

style.md
```diff
@@ -1557,7 +1557,237 @@ type Client struct {
...
+are either (a) exported or (b) unexported but conventionally satisfy public
+interfaces, especially if the embedding type can be trivially type-converted to
+an interface that the embedded type satsfies.
+
+#### Embedding Should Not Affect Useful Zero Values
```
prashantv: Big +1, this is great

style.md
```diff
@@ -1557,7 +1557,237 @@ type Client struct {
...
+
+</td></tr>
+</tbody></table>
+
+#### Embedding Should Not Change Type Semantics
+
+An embedded (inner) type should not add semantically unrelated or extraneous
+methods or fields to the embedding (outer) type's exported API, nor should it
+add methods or fields that imply that the type behaves other than explicitly
+defined. This applies both to newly-available (via embedding) and "shadowed"
```
prashantv: earlier we said never shadow, so i don't know if we should be talking about the shadow case here

style.md
```diff
@@ -1557,7 +1557,237 @@ type Client struct {
...
+methods and fields.
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+type Car struct {
```
prashantv: i think this example has a lot going on:
 - multiple embeds, neither of which is related and leaks unnecessary details
 - shadowing with a different intent

Can we distill this a little -- maybe just a `Car` type with a mutex?

style.md
```diff
@@ -1557,7 +1557,237 @@ type Client struct {
...
+    time.Ticker   // Bad: non-canonical
+}
+```
+
+</td><td>
+
+```go
+type Foo struct {
+    bytes.Buffer
+    *time.Ticker
```
prashantv: hmm I'm not sure about this recommendation, since `time.Ticker` is probably fine if we always use `*Foo` everywhere?

style.md
mway: Yeah, I think "never" is too strong a word. Some of the common workarounds for embedding where we'd essentially be shadowing are things like

```go
type Foo struct{}
func (f Foo) Bar(...) ... {}

type MultiFoo []Foo

func (mf MultiFoo) Bar(...) ... {
  for _, f := range mf {
    collect(f.Bar())
  }
}
```

etc. I guess a good litmus test would be something like, "should not prevent the underlying method from being called normally" or something to that effect - we're basically trying to argue that it's okay to augment functions by hijacking them, but while not opening the door for "shadowing is universally OK".

I'll take a stab at how to better phrase this, but if you have an idea, let me know!

style.md
mway: The unfortunate reality is probably "it depends":

- Is the zero type already useful? If yes, and the unexported type does not have a useful (and in this case, zero-touch) zero value, then that's bad.
- If the type now requires complex/opaque construction as a result of embedding the unexported type, then that's bad.
- If the embedded type's own fields are exported, that's bad.

I think if we established a basic guideline of "An embedded type must not impede existing trivial use of the embedding type" (i.e. to go from trivially constructable/copyable to not), that might be enough to discourage misuse while allowing for design choices that we don't have the ability to foresee.

This is all with the understanding that, in most cases, simply not embedding the type is the right answer. So maybe that's the right guidance:

> By default, embedding unexported types should be avoided. Embedded types must also not make the embedding type's zero value less useful, or make its construction opaque to the user.

Thoughts?

style.md
mway: It's fine if `*Foo` is the canonically-constructed version of "a Foo". If the constructor returns `Foo` instead of `*Foo`, that's probably less okay.

I need to rework these examples - agreed that they're not great.

style.md
mway: Yeah, I think "never" is too strong a word. Some of the common workarounds for embedding where we'd essentially be shadowing are things like

```go
type Foo struct{}
func (f Foo) Bar(...) ... {}

type MultiFoo []Foo

func (mf MultiFoo) Bar(...) ... {
  for _, f := range mf {
    collect(f.Bar())
  }
}
```

etc. I guess a good litmus test would be something like, "should not prevent the underlying method from being called normally" or something to that effect - we're basically trying to argue that it's okay to augment functions by hijacking them, but while not opening the door for "shadowing is universally OK".

I'll take a stab at how to better phrase this, but if you have an idea, let me know!

PR #55, opened 2019-10-20
Add guidance for initializing maps

### LINKED ISSUE #44: What is the preferred way to instantiate an empty map?
a nil slice is a valid slice and there's no problem doing `append(nilSlice, elem)`. However, a nil map isn't and it has to be instantiated before further use.

How do you recommend to instantiate an empty map? `m := map[type]type{}`  or  `m := make(map[type]type)` or something else?

mway:
Fixes #44.

### DIFF
--- a/style.md
+++ b/style.md
@@ -2012,12 +2012,12 @@ for i := 0; i < 10000; i++ {
 <tr><td>
 
 The footprint of `m` is not well-understood at initialization time,
-and multiple allocations may take place as part of adding elements.
+and allocations are deferred until elements are added.
 
 </td><td>
 
 The footprint of `m` is better understood at initialization time,
-and fewer allocations are incurred when adding within its capacity.
+and allocations occur up front.
 
 </td></tr>
 </tbody></table>
@@ -2051,17 +2051,9 @@ m := map[T1]T2{
 </tbody></table>
 
 
-While it's impractical to prescribe a universal strategy for map initialization,
-there are some basic rules of thumb you can use here.
-
-- Use `make()` when...
-  - an empty map is being initialized
-  - elements must be added programatically, versus as part of a literal
-  - when a size hint is available
-  - when the rest of the codebase predominantly uses `make()`
-- Use type literals when...
-  - a fixed set of elements is being added at initialization time
-  - when the rest of the codebase predominantly uses literals
+The basic rule of thumb is to use map literals when adding a fixed set of
+elements at initialization time, otherwise use `make` (and specify a size hint
+if available).
 
 There are more nuances than outlined in this guide; when in doubt,
 [be consistent](#be-consistent).
--- a/style.md
+++ b/style.md
@@ -2063,9 +2063,6 @@ The basic rule of thumb is to use map literals when adding a fixed set of
 elements at initialization time, otherwise use `make` (and specify a size hint
 if available).
 
-There are more nuances than outlined in this guide; when in doubt,
-[be consistent](#be-consistent).
-
 ### Format Strings outside Printf
 
 If you declare format strings for `Printf`-style functions outside a string
--- a/style.md
+++ b/style.md
@@ -1968,6 +1968,60 @@ Declaration and initialization are visually distinct.
 </td></tr>
 </tbody></table>
 
+Where possible, provide capacity hints when initializing
+maps with `make()`.
+
+```go
+make(map[T1]T2, hint)
+```
+
+While Go's runtime map implementation only acts upon capacity
+hints in certain situations (see
+[`runtime.makemap()`](https://golang.org/src/runtime/map.go)
+for more information), this can also be a helpful cue when
+reading source code.
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+m := make(map[T1]T2)
+
+// later...
+
+for i := 0; i < 10000; i++ {
+  // add element to m
+)
+```
+
+</td><td>
+
+```go
+m := make(map[T1]T2, 10000)
+
+// later...
+
+for i := 0; i < 10000; i++ {
+  // add element to m
+}
+```
+
+</td></tr>
+<tr><td>
+
+The footprint of `m` is not well-understood at initialization time,
+and multiple allocations may take place as part of adding elements.
+
+</td><td>
+
+The footprint of `m` is better understood at initialization time,
+and fewer allocations are incurred when adding within its capacity.
+
+</td></tr>
+</tbody></table>
+
 On the other hand, if the map holds a fixed list of elements,
 use map literals to initialize the map.
 
--- a/style.md
+++ b/style.md
@@ -1959,7 +1959,7 @@ Declaration and initialization are visually similar.
 
 </td><td>
 
-Declaration and initialization are more visually distinct.
+Declaration and initialization are visually distinct.
 
 </td></tr>
 </tbody></table>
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -1923,6 +1924,112 @@ sptr := &T{Name: "bar"}
...
+
+</td></tr>
+</tbody></table>
+
+While it's impractical to prescribe a universal strategy for map and slice
+initialization, there are some basic rules of thumb for when to use `make()`
+versus type literals:
+
+- Use `make(map[T1]T2)` or `make([]T, cap)` when...
+  - an empty map or slice is being initialized
```
abhinav: Oh I thought we were talking only about maps? For empty slices where the capacity is not known, `var x []Foo` is preferred because it's zero-cost. See also [CodeReviewComments#declaring-empty-slices](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

(Also, it would be `make([]T, 0, cap)`.)

style.md
```diff
@@ -1923,6 +1924,112 @@ sptr := &T{Name: "bar"}
...
+  - elements must be added programatically, versus as part of a literal
+  - [maps only] when it is optimal to pre-allocate storage (based on an
```
abhinav: I wouldn't point people to the map implementation. We can just suggest adding a size hint when available.

mway: sure, that sounds good - mostly copy/pasted this from the related issue.

style.md
mway: yeah, you're right. originally there was a reason to discuss both, but agreed that nil slices are preferable in that case. i'll just revert this to maps for now.

abhinav:
Proposing changes to this PR in #57.

style.md
```diff
@@ -1923,6 +1924,90 @@ sptr := &T{Name: "bar"}
...
+  - an empty map is being initialized
+  - elements must be added programatically, versus as part of a literal
+  - when a size hint is available
+  - when the rest of the codebase predominantly uses `make()`
+- Use type literals when...
+  - a fixed set of elements is being added at initialization time
+  - when the rest of the codebase predominantly uses literals
+
+There are more nuances than outlined in this guide; when in doubt,
+[be consistent](#be-consistent).
```
prashantv: This makes sense, but is also quite verbose. I'd prefer to keep recommendations short and succinct so it's easier to parse all the rules. Since we'll have be-consistent as a top-level guideline, I don't think we have to repeat it here.

I think this can be summarized as
```
Use map literals when a fixed set of elements is added at initialization time, otherwise use `make` and specify a size hint if available.

Comment (mway) [reply style.md]: modulo adding "when initializing an empty map", sgtm!

File: style.md [hunk 4]
```diff
@@ -1923,6 +1924,90 @@ sptr := &T{Name: "bar"}
 </td></tr>
 </tbody></table>
 
+### Initializing Maps
+
+Prefer `make(..)` for empty maps, and maps populated
```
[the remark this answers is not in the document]
mway: The reason I'd originally included a reference to the runtime hashmap implementation is because it's not obvious that size hints actually do not do anything unless the hint is large enough to require multiple buckets - which is different from slice allocation, which *does* allocate that memory based directly on the capacity.

I think it's fine to include a blanket recommendation to provide capacity, but it should probably also include a caveat that essentially says "just because you pass a hint doesn't necessarily mean that anything will be allocated".

What do you think?

style.md
```diff
@@ -1923,6 +1924,90 @@ sptr := &T{Name: "bar"}
...
+programmatically. This makes map initialization visually
+distinct from declaration, and it makes it easy to add size
```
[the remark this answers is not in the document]
mway: I went down a similar path before pushing the initial PR and decided not to - the tl;dr being that we should probably have some baseline expectation of the audience's familiarity with Go, particularly the behavior of builtin types.

Totally up to revisit that assumption, or for us to officially set a baseline of expected knowledge of the language - e.g. "for Go beginners", "for intermediate Go users", "for advanced Go users", "for anyone with some experience writing software", etc. What do you think?

style.md
```diff
@@ -1923,6 +1924,148 @@ sptr := &T{Name: "bar"}
...
+The footprint of `m` is not well-understood at initialization time,
+and multiple allocations may take place as part of adding elements.
+
+</td><td>
+
+The footprint of `m` is better understood at initialization time,
+and fewer allocations are incurred when adding within its capacity.
+
+</td></tr>
+</tbody></table>
```
mway: @prashantv thoughts?

style.md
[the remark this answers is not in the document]
prashantv: The optimizations for small size hints are interesting, but I think they end up being an implementation detail that the user shouldn't worry about -- if they have a size hint, they should pass it in.

The main difference between smaller/larger size hints seems to be:
 - For larger sizes, the `*hmap` and the buckets are pre-allocated.
 - For smaller maps, the `*hmap` is allocated, and the first assignment causes the bucket to be created.

In both cases, the total number of allocations is the same, but where the allocation happens can move around. It can definitely be a little surprising if a size hint was passed in and there still was an allocation on an assignment, but I don't think it's worth calling out specifically here.

style.md
```diff
@@ -1923,6 +1924,140 @@ sptr := &T{Name: "bar"}
...
+
+
+The basic rule of thumb is to use map literals when adding a fixed set of
+elements at initialization time, otherwise use `make` (and specify a size hint
+if available).
+
+There are more nuances than outlined in this guide; when in doubt,
```
prashantv: optional: OK to have this, but think it could be dropped without much of an impact, since we've called out "basic rule of thumb" above.

style.md
prashantv: I like the recommendation, but think it makes this guideline a little too long. A suggestion:
 - drop the performance-related text here
 - add a separate recommendation in the `Performance` section about specifying size hints with most of this text
 - reference that section here when we suggest the user should specify a size hint

The separate section can also be used for slices.

What do you think?

mway: Agreed, sounds good!

style.md
mway: Probably right. Will remove.

style.md
```diff
@@ -1047,6 +1049,62 @@ BenchmarkGood-4  500000000   3.25 ns/op
...
+
+</td></tr>
+<tr><td>
+
+Bucket allocations are deferred until elements are added.
+
+</td><td>
+
+Bucket allocations occur at initialization time
+(for sufficiently large capacity hints).
```
prashantv: These may be better as `Less allocations at assignment time` vs `More allocations at assignment time`?

style.md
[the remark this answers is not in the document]
prashantv: Good question, I haven't really thought about a specific audience other than Go engineers at Uber. I'd say mostly beginner/intermediate?

In this case, it might be easy to miss that declaration = `nil` map, but since we call out the `m2 will panic on writes` below, it might be less important to call this out here.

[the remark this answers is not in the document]
mway: Yeah, it'd definitely be easier to expand on this as we find that folks are confused by it.

style.md
```diff
@@ -1047,6 +1049,62 @@ BenchmarkGood-4  500000000   3.25 ns/op
...
+```go
+make(map[T1]T2, hint)
+```
+
+While Go's runtime map implementation only acts upon capacity
+hints in certain situations (see
+[`runtime.makemap()`](https://golang.org/src/runtime/map.go)
+for more information), providing a hint to `make()` can
+cause allocations to occur at map initialization time rather
+than periodically as elements are added to the map.
```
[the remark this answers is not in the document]
mway: I like your summary better - thanks!

PR #73, opened 2019-11-26
Port over the section on mutable globals

nomis52:


style.md
```diff
@@ -965,6 +966,83 @@ func (f *foo) isRunning() bool {
...
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+// foo.go
+
+var _doFoo = foo
```
prashantv: can we make this example a little more realistic, maybe a `timeNow func() time.Duration` function, since that's commonly stubbed in tests

abhinav: Going to switch to time.Now for message signing.

prashantv: Thanks @abhinav  -- should we use dependency injection here rather than changing the unexported field directly? That would also go better with the recommendation.

abhinav: I think @nomis52 here was specifically going for the "set the field" case.
Proper dependency injection is preferable but for this example, we would then
have to take a position on a clock library.  Right now, the example takes the
form of a privately injected dependency.  That isn't perfect but it's better
than the global form.

I can make it a parameter of the function (whether as a func or an interface),
but I don't know if that's the best idea here since we probably wouldn't do
that in practice depending on how coupled time is to the contract of the
abstraction.

nomis52: I've realized we need a separate section on dependency injection. My position is that for unexported types it's ok to set fields directly.

For exported types we should be using the Options pattern.

Thoughts?

abhinav: That's reasonable. The options pattern does allow for private options that are available to tests only. In the interest of keeping this example simple, we can make the object private.

prashantv: Fair enough, it's simpler to set fields, so let's go with that, and separately explore recommendations for dependency injection.

abhinav:
Ready for one last look @prashantv @mway

style.md
```diff
@@ -965,6 +966,82 @@ func (f *foo) isRunning() bool {
 </td></tr>
 </tbody></table>
 
+### Avoid Mutable Globals
+
+Avoid mutating global variables, instead opting for dependency injection.
+This applies to function pointers as well as other kinds of values.
```
mway: nit: We might want to expand this to include "don't introduce mutable global variables" in addition to "don't mutate global variables". Without getting overly specific, this could look something like:

> Avoid mutating global variables, instead opting for dependency injection.
> This applies to any assignable values that could be used as global state,
> such as function pointers, strings, and other kinds of values. Errors are
> a notable exception to this guidance.
>
> Similarly, avoid introducing mutable global variables by using `const` where possible.

The last part, in particular, is probably worth its own quick section - i.e., `const` wherever possible.

mway: (also, should've clarified - I made some minor tweaks to reference "assignable values" given that function pointers are perhaps not representative enough of what could be considered "mutable global state", given how folks often use strings or scalars to serve as "static config" in addition to overriding global function vars.)

abhinav:
Going to merge as-is. @mway, we can add the "avoid introducing globals" separately.

PR #76, opened 2019-12-10
Guideline: Use "time" to handle time

abhinav:
This adds a new section recommending use of the time package when
dealing with time. It ports over recommendations for `time.Duration` and
`time.Time` from the internal guide, explaining that time is complicated
as rationale.

### DIFF
--- a/style.md
+++ b/style.md
@@ -565,6 +565,8 @@ helps deal with these incorrect assumptions in a safer, more accurate manner.
 
   [`"time"`]: https://golang.org/pkg/time/
 
+#### Use `time.Time` for instants of time
+
 Use [`time.Time`] when dealing with instants of time, and the methods on
 `time.Time` when comparing, adding, or subtracting time.
 
@@ -591,6 +593,8 @@ func isActive(now, start, stop time.Time) bool {
 
 </td></tr>
 </tbody></table>
+
+#### Use `time.Duration` for periods of time
 
 Use [`time.Duration`] when dealing with periods of time.
 
@@ -641,6 +645,8 @@ use [`Time.Add`].
 newDay := t.AddDate(0 /* years */, 0, /* months */, 1 /* days */)
 maybeNewDay := t.Add(24 * time.Hour)
 ```
+
+#### Use `time.Time` and `time.Duration` with external systems
 
 Use `time.Duration` and `time.Time` in interactions with external systems when
 possible. For example:
--- a/style.md
+++ b/style.md
@@ -652,15 +652,22 @@ maybeNewDay := t.Add(24 * time.Hour)
 Use `time.Duration` and `time.Time` in interactions with external systems when
 possible. For example:
 
-- Command-line flags: [`flag`] supports `time.Duration`
-- JSON: [`encoding/json`] supports `time.Time`
-- SQL: [`database/sql`] supports `time.Time`
-- YAML: [`gopkg.in/yaml.v2`] supports `time.Time` and `time.Duration`
+- Command-line flags: [`flag`] supports `time.Duration` via
+  [`time.ParseDuration`]
+- JSON: [`encoding/json`] supports encoding `time.Time` as an [RFC 3339]
+  string via its [`UnmarshalJSON` method]
+- SQL: [`database/sql`] supports converting `DATETIME` or `TIMESTAMP` columns
+  into `time.Time` and back if the underlying driver supports it
+- YAML: [`gopkg.in/yaml.v2`] supports `time.Time` as an [RFC 3339] string, and
+  `time.Duration` via [`time.ParseDuration`].
 
-  [`encoding/json`]: https://golang.org/pkg/encoding/json/
-  [`gopkg.in/yaml.v2`]: https://godoc.org/gopkg.in/yaml.v2
-  [`database/sql`]: https://golang.org/pkg/database/sql/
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -545,6 +546,102 @@ const (
 // LogToStdout=0, LogToFile=1, LogToRemote=2
 ```
 
+### Use time
```
prashantv: Maybe `Use package time for time related types and operations`

style.md
```diff
@@ -545,6 +546,102 @@ const (
...
+
+Time is complicated. Incorrect assumptions often made about time include the
```
prashantv: I think there's a few parts to the guidance that we'd like:
 - Don't try to do time related calculations yourself, since it's easy to get wrong
 - Use the time related types where possible. E.g., it's supported in flags, YAML, etc.
 - If you need to use int (e.g., interacting with some external system over JSON), then the field should be named with the unit `fooSeconds`.

I'm not sure if they all belong in a single guideline

style.md
```diff
@@ -545,6 +546,102 @@ const (
...
+following.
+
+1. A day has 24 hours
+2. An hour has 60 minutes
+3. A week has 7 days
+4. A year has 365 days
+5. [And a lot more](https://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time)
```
prashantv: (just to confirm, go's time package does handle the cases above correctly right?)

abhinav: Yeah, it handles them correctly -- mostly. It doesn't support leap seconds per the documentation (see also golang/go#15247, golang/go#8728).

> The calendrical calculations always assume a Gregorian calendar, with no leap seconds.

style.md
abhinav: I would like to avoid making the titles too long. Is there something shorter we can put here?
How about, `Use package time to handle time` or  `Use "time" to handle time`?

prashantv: I like `Use "time" to handle time`

style.md
prashantv: Ah, do you think we should make a note of that here?

style.md
abhinav: This guidance only covers the first one of those. The second advantage
comes implicitly from using the correct types. Do you want me to
include that as an advantage here?

RE: the third: Instead of making it a time thing, that could be a more
general naming advice around including units in variable names. What do
you think?

style.md
abhinav: Yeah, I'll add a thing.

style.md
prashantv: I think 2nd is worth explicitly calling out, since there are places where you don't necessarily do calculations (e.g., passing in a timeout to some call) but you should still use the type in flags etc.

I like the 3rd being a separate guidance, but we may want to briefly mention or reference here

abhinav: Fair. Added a specific subsection about external systems, recommending time.Time and time.Duration where possible, and int/float with units for time.Duration, strings in RFC 3339 format for time.Time, when not possible.

https://github.com/uber-go/guide/blob/abg/time/style.md#use-timetime-and-timeduration-with-external-systems

style.md
```diff
@@ -546,6 +547,167 @@ const (
...
+maybeNewDay := t.Add(24 * time.Hour)
+```
+
+#### Use `time.Time` and `time.Duration` with external systems
+
+Use `time.Duration` and `time.Time` in interactions with external systems when
+possible. For example:
+
+- Command-line flags: [`flag`] supports `time.Duration`
+- JSON: [`encoding/json`] supports `time.Time`
```
prashantv: optional: It wasn't clear how it supported this, so I wrote a quick check
https://play.golang.org/p/gkg9i7Lebo8

Should we mention what "support" means? Can do one or more of the following:
 - Mention that it encodes as a RFC3399 string
 - Link to [specific docs](https://golang.org/pkg/time/#Time.MarshalJSON) for `MarshalJSON` that say "The time is a quoted string in RFC 3339 format, with sub-second precision added if present."
 - link to an example like the playground link

We should do the same for yaml/sql etc too ideally

style.md
```diff
@@ -546,6 +547,167 @@ const (
...
+For example, since `encoding/json` does not support `time.Duration`, the unit
+is included in the name of the field.
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+// {"interval": 2}
```
prashantv: since we don't have `json` tags, I think the output doesn't match what it really would look like -- suggest adding the json tag to make to reinforce the "json marshalling" in this example

style.md
```diff
@@ -546,6 +547,167 @@ const (
...
+  IntervalMillis int
+}
+```
+
+</td></tr>
+</tbody></table>
+
+When it is not possible to use `time.Time` in these interactions, unless an
+alternative is agreed upon, use `string` and format timestamps as defined in
+[RFC 3339]. This format is used by default by [`Time.UnmarshalText`].
```
prashantv: should we also mention the time constant `time.RFC3339`

style.md
abhinav: Makes sense. I'll elaborate on the "supports" for all cases.

style.md
abhinav: 👍  linking to constant for use in time.Parse and Time.Format.

PR #77, opened 2019-12-11
Functional Options: Don't recommend closures

### LINKED ISSUE #74: Guideline: For functional options, don't recommend closures
Our recommendation right now includes this pattern,

```
type Option interface{ apply(*options) }

type optionFunc func(*options)
```

And then all options are implemented using `optionFunc` and closures.

```
func WithTimeout(d time.Duration) Option {
  return optionFunc(func(o *options) {
    o.Timeout = d
  })
}
```

This is not the pattern we should be encouraging because it makes `Option`s
too opaque. We cannot print them or compare them, or introspect values inside
them.

We should instead recommend declaring a new type for each option.

```
type timeoutOption time.Duration

func WithTimeout(t time.Duration) Option {
  return timeoutOption(t)
}

func (o timeoutOption) apply(opts *options) {
  opts.Timeout = time.Duration(o)
}

// And

type loggerOption struct{ L *zap.Logger }

func WithLogger(l *zap.Logger) Option {
  return loggerOption{L: l}
}

func (o *loggerOption) apply(opts *options) {
  o.Logger = o.L
}
```

abhinav:
Per #74, this changes the recommendation for functional options to
suggest manual interface implementations instead of closures.

To make the example more demonstrative, I had to add a new option which
hurt the overall readability of the side-by-side comparison. This lead
to a minor reorganization: side-by-side includes user-facing API and
usage only, and implementation is demonstrated separately in its own
paragraph.

Resolves #74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 2019-12-11
+# 2019-12-17
 
 - Functional Options: Recommend struct implementations of `Option` interface
   instead of capturing values with a closure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2019-12-11
+
+- Functional Options: Recommend individual implementations of `Option`
+  interface instead of capturing values with a closure.
+
 # 2019-11-26
 
 - Add guidance against mutating global variables.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # 2019-12-11
 
-- Functional Options: Recommend individual implementations of `Option`
-  interface instead of capturing values with a closure.
+- Functional Options: Recommend struct implementations of `Option` interface
+  instead of capturing values with a closure.
 
 # 2019-11-26
 
--- a/style.md
+++ b/style.md
@@ -2383,7 +2383,7 @@ db.Open(addr, false /* cache */, log)
 
 </td><td>
 
-Options must be provided only if needed.
+Options are provided only if needed.
 
 ```go
 db.Open(addr)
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 2019-12-11
+
+- Functional Options: Recommend individual implementations of `Option`
```
prashantv: I think `Recommend struct implementations` would be more accurate (some options may reuse the same struct)

style.md
```diff
@@ -2333,62 +2333,116 @@ more arguments on those functions.
...
+```go
+db.Open(addr, db.DefaultCache, zap.NewNop())
+db.Open(addr, db.DefaultCache, log)
+db.Open(addr, false /* cache */, zap.NewNop())
+db.Open(addr, false /* cache */, log)
 ```
 
 </td><td>
 
+Options must be provided only if needed.
```
prashantv: "must be provided" sounds a little odd, maybe just "Options are only provided if needed"

style.md
```diff
@@ -2397,21 +2451,11 @@ func Connect(
...
-  db.WithCaching(false),
-  db.WithTimeout(newTimeout),
-)
 ```
 
-</td></tr>
-</tbody></table>
+Note that there's a method of implementing this pattern with closures but we
+believe that the pattern above provides more flexibility for authors and is
+easier to debug for users.
```
prashantv: can we provide more details here?
 - loss of comparisons (also makes mocks much harder)
 - loss of string methods for pretty-printing

style.md
abhinav: oops, yeah

PR #78, opened 2019-12-26
Guideline: avoid memory leak in http requests

lucassscaravelli:


style.md
```diff
@@ -1182,6 +1183,72 @@ allocations at assignment time.
 </td></tr>
 </tbody></table>
 
+### Avoid memory leak in http requests
+
+In a program that makes HTTP requests, it's important to close the body request. If not,
+your application will stack memory consume, resulting in a memory leak problem.
+
+Another good practice you must do, it's read the data body even the data is not important,
+it avoid memory leak in http requests.
```
mlaventure:
```suggestion
Another good practice you must do, is to read the data body even if the data is not important,
it avoids memory leak in http requests.
```

mway:
Hey @lucassscaravelli! Thanks for the submission - we discussed this a bit internally and, while definitely correct, feel that this guidance is specific to a particular use case versus general, language-level guidance, so probably doesn't make sense as part of the guide.

Please let us know if you have other guidance you think should be included!

PR #84, opened 2020-01-23
Avoid embedding types in public structs

kriskowal:
This change introduces a section about avoiding the use of type embedding because it limits the evolution of the container type.

### DIFF
--- a/style.md
+++ b/style.md
@@ -67,6 +67,7 @@ row before the </tbody></table> line.
   - [Don't Panic](#dont-panic)
   - [Use go.uber.org/atomic](#use-gouberorgatomic)
   - [Avoid Mutable Globals](#avoid-mutable-globals)
+  - [Embedding Considered Harmful](#embedding-considered-harmful)
 - [Performance](#performance)
   - [Prefer strconv over fmt](#prefer-strconv-over-fmt)
   - [Avoid string-to-byte conversion](#avoid-string-to-byte-conversion)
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -1042,6 +1043,136 @@ func TestSigner(t *testing.T) {
 </td></tr>
 </tbody></table>
 
+### Embedding considered harmful
```
mway: The first line of this section - "Avoid embedding types in public structs" - is probably a better title. Let's use that instead.

style.md
```diff
@@ -1042,6 +1043,136 @@ func TestSigner(t *testing.T) {
...
+
+Avoid embedding types in public structs.
+These obscure documentation and leak a detail about your implementation that
```
mway: nit: Leaking details >> inhibiting type evolution > obscured documentation, IMO - so maybe we can rewrite this as:

> Embedding can leak implementation details, make type evolution difficult, or obscure type documentation.

(or something along those lines)

style.md
```diff
@@ -1042,6 +1043,136 @@ func TestSigner(t *testing.T) {
...
+func (l *ConceteList) Remove(e Entity) {
+    return l.list.Remove(e)
+}
+```
+
+</td></tr>
+</tbody></table>
+
+Although writing these delegate methods is tedious, it leaves more
+opportunities for change open and also eliminates indirection for discovering
```
mway: nit: I'd probably mention that this avoids leaking implementation details again.

style.md
```diff
@@ -1042,6 +1043,129 @@ func TestSigner(t *testing.T) {
...
 
+### Avoid Embedding Types in Public Structs
+
+These leak implementation details, inhibit type evolution, and obscure
+documentation.
+
+Assuming you have implemented a variety of list types using a shared
+`AbstractList`, avoid embedding the `AbstractList` in your concrete list
+implementations.
+Instead, hand-write only the methods to your concrete list that they will
```
mway: ultra nit:

> Instead, hand-write only the methods to your concrete list that will be delegated to the abstract list.

style.md
```diff
@@ -1042,6 +1043,129 @@ func TestSigner(t *testing.T) {
...
+delegate to the abstract list.
+
+```go
+type AbstractList struct {}
+
+// Add adds an entity to the list.
+func (l *AbstractList) Add(e Entity) {
+    // ...
+}
```
mway: nit: [We're using 2-space indentation](https://raw.githubusercontent.com/uber-go/guide/master/style.md) for code examples.

style.md
kriskowal: call and raise
> Instead, hand-write only the methods to your concrete list that will delegate to the abstract list.

style.md
```diff
@@ -1042,6 +1043,129 @@ func TestSigner(t *testing.T) {
 </td></tr>
 </tbody></table>
 
+### Avoid Embedding Types in Public Structs
+
+These leak implementation details, inhibit type evolution, and obscure
```
prashantv: (optional)

s/These/Embedded types

Issue #25, opened 2019-10-12
Consider adding "var _ Interface = (*Type)(nil)" to the list of recommendations


shyiko:
This is a relatively common Go idiom equivalent to Java's "implements", Rust's "impl ... for", etc.
It
- asserts that types implement expected set of interfaces (declaration of intent), guarding against breakage caused by method signature changes
- improves readability by making things explicit & acting as an additional form of documentation
- helps with code navigation/generation (e.g. in GoLand Alt+Insert on `(*Type)(nil)` gives `Implement Missing Methods` option).

<table>
<thead><tr><th>Bad</th><th>Good</th></tr></thead>
<tbody>
<tr><td>

```go
type A struct{}

func New() A {
	return A{}
}

func (a A) Read(p []byte) (n int, err error) {
	panic("implement me")
}

func (a A) Close() error {
	panic("implement me")
}
```

</td><td>

```go
type A struct{}

var _ io.ReadCloser = (*A)(nil)

func New() A {
	return A{}
}

func (a A) Read(p []byte) (n int, err error) {
	panic("implement me")
}

func (a A) Close() error {
	panic("implement me")
}
```

</td></tr>
</tbody></table>

I'd be happy to submit a PR.

prashantv:
+1, if a type is expected to satisfy an interface, and we don't already verify this, then the `var _ <interface> = <type instantiation>` makes sense.

An alternate is to make the `New` function return the interface type directly, in which case the variable declaration is unnecessary.

abhinav:
+1 on the `var _` recommendation.

Having `New` return the interface would go against the "return structs" adage,
so I'd like to avoid that for now.

prashantv:
Good point.

There are cases when the compiler will already have the check, e.g., a custom implementation of [`net.Listener`](https://golang.org/pkg/net/#Listener) returns a `net.Conn` from `Accept`, so it's technically unnecessary to add another check if a custom connection type is returned.

That said, it seems simpler to just add the recommendation, and not worry about the corner cases.

abhinav:
Landed

PeterlitsZo:
If the interface is defined in a package `foo`, and I try to let struct `BarStruct` implement it in another package `bar`.

Now I invoke `NewBarStruct` in package `foo`, so now:

- the package `foo` depends on package `bar` (to invoke `NewBarStruct` function),
- the package `bar` depends on package `foo` (to use statement `var _ (foo.FooInterface) = BarStruct{}`)

[...]

PR #86, opened 2020-02-16
Guidelines: Verify Interface Compliance

abhinav:
This adds advice on verifying interface compliance with
`var _ Interface = (*Foo)(nil)`.

As part of this guideline, we have to cover that if the receiver is on
the value type, the right side should be the zero value. Because of
that, the guideline is placed right before the section where we talk
about receiver types for interfaces.

Resovles #25

---

[Preview](https://github.com/uber-go/guide/blob/abg/interface-compliance/style.md#verify-interface-compliance)

### DIFF
--- a/style.md
+++ b/style.md
@@ -55,6 +55,7 @@ row before the </tbody></table> line.
 - [Introduction](#introduction)
 - [Guidelines](#guidelines)
   - [Pointers to Interfaces](#pointers-to-interfaces)
+  - [Verify Interface Compliance](#verify-interface-compliance)
   - [Receivers and Interfaces](#receivers-and-interfaces)
   - [Zero-value Mutexes are Valid](#zero-value-mutexes-are-valid)
   - [Copy Slices and Maps at Boundaries](#copy-slices-and-maps-at-boundaries)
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -151,6 +152,65 @@ An interface is two fields:
...
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+type Handler struct{ /* ... */ }
+
+
+
```
mway: nit: I think we can condense this whitespace

style.md
```diff
@@ -151,6 +152,65 @@ An interface is two fields:
...
+For types that are known or required to implement specific interfaces, add
+checks to verify interface compliance at compile time.
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+type Handler struct{ /* ... */ }
```
mway: ultra style nit: either break this into multiple lines (as it would be defined outside of an example) or add a space before the opening brace (here and below for the adjacent `Handler` definition)

style.md
```diff
@@ -151,6 +152,65 @@ An interface is two fields:
...
+</td></tr>
+</tbody></table>
+
+The statement `var _ http.Handler = (*Handler)(nil)` will fail to compile if
+`*Handler` ever stops matching the `http.Handler` interface.
+
+Note that the interface was implemented on the pointer type `*Handler` above,
+which meant use of `(*Handler)(nil)` on the right side of `=`. If the
+interface was implemented on the value type, the right side would use the zero
+value of that type.
```
mway: since we're being pedantic - the zero value of a pointer type is `nil`, so the last sentence is true regardless of whether a pointer or value type is being asserted as implementing an interface - so it might be worth just standardizing this to something like:

> The right hand side of the assignment should be the zero value of the type being asserted - for pointers, this is `nil`; for a value type `T`, this is `T{}`: [optional side by side example]

I do think that a struct would be more illustrative than an integer type, though. Or both would be fine, but I'd at least include a struct since that's more syntactically confusing.

style.md
abhinav: The intent was for it to align with the addition of `var _ ...` on the right side of the table.

image

I can get rid of it if the extraneous whitespace isn't worth it.

mway: ah, I was only looking at this in diff form, not in rendered form. makes sense.

style.md
```diff
@@ -151,6 +152,65 @@ An interface is two fields:
 If you want interface methods to modify the underlying data, you must use a
 pointer.
 
+### Verify Interface Compliance
+
+For types that are known or required to implement specific interfaces, add
+checks to verify interface compliance at compile time.
```
prashantv: Should this still be done if the constructor for the type has a signature that returns the interface, but the implementation returns the struct? 

Does it also mean that every type that implements `Stringer` should do this?

abhinav: Per https://github.com/uber-go/guide/issues/25#issuecomment-541377764, I figured we were going to add the recommendation without covering things like "it's not necessary if a constructor or another function returns it."

RE: stringer: Switched away from it. It's usually unnecessary to verify Stringer compliance but sometimes it's part of the contract, in which case you do want it.

I previously used phrasing that asks this only for "types that are known or required to implement specific interfaces." We can narrow it down to "types that are required to implement specific interfaces," making this a requirement only if it's part of the contract of a type.

abhinav: Oh and if we add the 'exported' qualifier, it covers the case of constructors returning unexported implementations.

abhinav:
Reworded to "where appropriate" with examples.

PR #91, opened 2020-05-29
Add reminder that nil slices are not always the same as empty slices

### LINKED ISSUE #90: "nil is a valid slice" issue
Actually []int{} and nil are not equivalent. Replacement of one to another may cause broken api.

Example:
https://play.golang.org/p/MojffFnM8Gn

There should be warning about that.

mway:
Fixes #90.


abhinav:
LGTM

PR #94, opened 2020-06-04
Add guidance on avoiding init()

### LINKED ISSUE #72: Guideline: Avoid init
Creating this to track adding a general guideline around use or misuse
of `init()`.

We can formulate the full guideline in this issue, but the first piece
of that can be based on the following general advice:

> If it doesn't behave exactly the same every time it's called,
regardless of environment, arguments, directory, etc., then it most
definitely does not belong in `init()`.

I think our guideline should distill that and at least the following
points:

- Avoid `init()` if you can
- Pre-computation is okay as long as it's deterministic
- Don't do any form of IO from `init()`
- Don't rely on mutable global state in `init()` (command line
  arguments, environment variables, current directory, etc.)

mway:
Fixes #72.

### DIFF
--- a/style.md
+++ b/style.md
@@ -71,6 +71,7 @@ row before the </tbody></table> line.
   - [Avoid Mutable Globals](#avoid-mutable-globals)
   - [Avoid Embedding Types in Public Structs](#avoid-embedding-types-in-public-structs)
   - [Avoid Using Built-In Names](#avoid-using-built-in-names)
+  - [Avoid `init()`](#avoid-init)
 - [Performance](#performance)
   - [Prefer strconv over fmt](#prefer-strconv-over-fmt)
   - [Avoid string-to-byte conversion](#avoid-string-to-byte-conversion)
@@ -1514,6 +1515,114 @@ func (f Foo) String() string {
 Note that the compiler will not generate errors when using predeclared
 identifiers, but tools such as `go vet` should correctly point out these and
 other cases of shadowing.
+
+### Avoid `init()`
+
+Avoid `init()` where possible. At a minimum, all code within `init()` should:
+
+1. Be completely deterministic, regardless of program environment or invocation.
+2. Be completely self-contained within the program.
+3. Not depend on side-effects of other `init()` functions.
+4. Be able to run in any order. While `init()` ordering is well-known, code can
+   change, and thus `init()` functions that rely on specific sequencing make
+   code brittle and error-prone.
+5. Not access or manipulate global or environment state, such as machine
+   information, environment variables, working directory, program
+   arguments/inputs, etc.
+6. Not perform I/O, including both filesystem, network, and system calls.
+
+Code that cannot satisfy these requirements likely belongs as a helper to be
+called as part of `main()` (or elsewhere in a program's lifecycle), or be
+written as part of `main()` itself. In particular, libraries that are intended
+to be used by other programs should take special care to be completely
+deterministic and not perform "init magic".
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+type Foo struct {
+    // ...
+}
+
+var _defaultFoo Foo
+
+func init() {
+    _defaultFoo = Foo{
+        // ...
+    }
+}
+```
+
+</td><td>
+
+```go
+var _defaultFoo = Foo{
+    // ...
+}
+```
+
+</td></tr>
+<tr><td>
+
+```go
+type Config struct {
+    // ...
+}
+
+var _config Config
+
+func init() {
+    // Bad: based on current directory
+    cwd, _ := os.Getwd()
+
+    // Bad: I/O
+    raw, _ := ioutil.ReadFile(
+        path.Join(cwd, "config", "config.yaml"),
+    )
+
+    yaml.Unmarshal(raw, &_config)
+}
+```
+
+</td><td>
+
+```go
+type Config struct {
+    // ...
+}
+
+func loadConfig() Config {
+    cwd, err := os.Getwd()
+    // handle err
+
+    raw, err := ioutil.ReadFile(
+        path.Join(cwd, "config", "config.yaml"),
+    )
+    // handle err
+
+    var config Config
+    yaml.Unmarshal(raw, &config)
+
+    return config
+}
+```
+
+</td></tr>
+</tbody></table>
+
+Considering the above, some situations in which `init()` may be preferable or
+necessary might include:
+
+- Complex expressions that cannot be represented as single assignments.
+- Seeding or constructing default helpers or state, such as `math/rand.Seed`.
+- Pluggable hooks, such as `database/sql` dialects, encoding type registries, etc.
+- Optimizations to [Google Cloud Functions] and other forms of deterministic
+  precomputation.
+
+  [Google Cloud Functions]: https://cloud.google.com/functions/docs/bestpractices/tips#use_global_variables_to_reuse_objects_in_future_invocations
 
 ## Performance
 
--- a/style.md
+++ b/style.md
@@ -1562,6 +1562,16 @@ func init() {
 var _defaultFoo = Foo{
     // ...
 }
+
+// or, better, for testability:
+
+var _defaultFoo = defaultFoo()
+
+func defaultFoo() Foo {
+    return Foo{
+        // ...
+    }
+}
 ```
 
 </td></tr>
@@ -1617,7 +1627,6 @@ Considering the above, some situations in which `init()` may be preferable or
 necessary might include:
 
 - Complex expressions that cannot be represented as single assignments.
-- Seeding or constructing default helpers or state, such as `math/rand.Seed`.
 - Pluggable hooks, such as `database/sql` dialects, encoding type registries, etc.
 - Optimizations to [Google Cloud Functions] and other forms of deterministic
   precomputation.
--- a/style.md
+++ b/style.md
@@ -1518,18 +1518,18 @@ other cases of shadowing.
 
 ### Avoid `init()`
 
-Avoid `init()` where possible. At a minimum, all code within `init()` should:
+Avoid `init()` where possible. When `init()` is unavoidable or desirable, code
+should attempt to:
 
 1. Be completely deterministic, regardless of program environment or invocation.
-2. Be completely self-contained within the program.
-3. Not depend on side-effects of other `init()` functions.
-4. Be able to run in any order. While `init()` ordering is well-known, code can
-   change, and thus `init()` functions that rely on specific sequencing make
-   code brittle and error-prone.
-5. Not access or manipulate global or environment state, such as machine
+2. Avoid depending on the ordering or side-effects of other `init()` functions.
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+   arguments/inputs, etc.
+6. Not perform I/O, including both filesystem, network, and system calls.
+
+Code that cannot satisfy these requirements likely belongs as a helper to be
+called as part of `main()` (or elsewhere in a program's lifecycle), or be
+written as part of `main()` itself. In particular, libraries that are intended
+to be used by other programs should take special care to be completely
+deterministic and not perform "init magic".
+
+<table>
```
mway, line 1540: There's probably room for 1-2 other (potentially more illustrative) examples; open to suggestions.

style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+}
+
+func loadConfig() Config {
+    cwd, _ := os.Getwd()
+    raw, _ := ioutil.ReadFile(
+        path.Join(cwd, "config", "config.yaml"),
+    )
+
+    var config Config
+    yaml.Unmarshal(raw, &config)
```
abhinav, line 1604: For the good and bad example in this case, I'm not 100% sure about this, but
it may be preferable to still include `err`s and gloss over error handling
with `// handle error` comments.

    cwd, err := os.Getwd()
    // handle error

style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+
+    return config
+}
+```
+
+</td></tr>
+</tbody></table>
+
+Notable exceptions to `init()` determinism guidance include situations such as
+seeding non-cryptographic random number generators used across a codebase:
```
abhinav, line 1614: Are we sure about this exception? It may be preferable for the `*rand.Rand` to be passed around explicitly, especially so that we can get determinism in tests of randomness.

style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+```
+
+Considering the above, some situations in which `init()` may be preferable or
+necessary might include:
+
+- To reduce or prevent user error
+- Complex expressions that cannot be represented as single assignments
+- Conditional assignments
+- Seeding or constructing default helpers or state, such as `math/rand.Seed`
+- Pluggable hooks, such as `database/sql` dialects, encoding type registries, etc
```
abhinav, line 1629: Another exception: Google Cloud Functions (https://github.com/uber-go/guide/issues/72#issuecomment-562329745). We don't need an example for that one; a mention should suffice.

(Probably AWS Lambda too, but I don't know that for sure so let's leave that out for now.)

style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+    rand.Seed(time.Now().UnixNano())
+}
+```
+
+Considering the above, some situations in which `init()` may be preferable or
+necessary might include:
+
+- To reduce or prevent user error
+- Complex expressions that cannot be represented as single assignments
+- Conditional assignments
```
abhinav, line 1627: what's this?

style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+```go
+func init() {
+    rand.Seed(time.Now().UnixNano())
+}
+```
+
+Considering the above, some situations in which `init()` may be preferable or
+necessary might include:
+
+- To reduce or prevent user error
```
abhinav, line 1625: This is unclear

style.md
```diff
@@ -1515,6 +1516,118 @@ Note that the compiler will not generate errors when using predeclared
...
+- Complex expressions that cannot be represented as single assignments
```
abhinav, line 1626: consistency nit: items in prior lists have been full sentences ending with periods. Why not the same here?

style.md
mway: Yeah, I indexed on brevity versus completeness since real estate is pretty limited here. I'll call out error handling, though.

style.md
mway: Good point about tests - I'll just omit this, I don't think there's a ton of value to delving too deep here.

style.md
mway: It seemed reasonable to call out conditional assignments like:

```go
var myVar T

func init() {
  if constexpr {
    myVar = <variant A>
  } else {
    myVar = <variant B>
  }
}
```

but if we feel it adds too much ambiguity, we can omit it.

style.md
mway: Yeah, I didn't really have a good concrete example offhand other than something about providing sane defaults (like in situations where folks don't control the zero values or there isn't a useful zero value). Could go either way on this one, really.

style.md
abhinav: Yeah, perhaps we should omit this for now. Without the explanation, one interpretation is that it's okay to do `if os.Getenv(...) == "something" { v = a } else { v = b }`, which violates the guidelines stated before for what's acceptable in init.

style.md
abhinav: We don't necessarily need examples for all exceptions, but for this one, it's
not clear to me how init would help here. We can drop it and add it back later
if it becomes more obvious with different phrasing/with an example.

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+</td><td>
+
+```go
+type Config struct {
+    // ...
+}
+
+func loadConfig() Config {
+    cwd, osErr := os.Getwd()
+    // handle osErr
```
abhinav, line 1599: Any reason not to just use err?

```suggestion
    cwd, err := os.Getwd()
    // handle err
```

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+
+    raw, readErr := ioutil.ReadFile(
+        path.Join(cwd, "config", "config.yaml"),
+    )
+    // handle readErr
```
abhinav, line 1604:
```suggestion
    raw, err := ioutil.ReadFile(
        path.Join(cwd, "config", "config.yaml"),
    )
    // handle err
```

style.md
mway: Hmm, I just re-ran without issues, but one of the lint configs I'd used recently had complained about shadowing/redefinition. Changed this to just be `err` for simplicity.

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
 identifiers, but tools such as `go vet` should correctly point out these and
 other cases of shadowing.
 
+### Avoid `init()`
+
+Avoid `init()` where possible. At a minimum, all code within `init()` should:
+
+1. Be completely deterministic, regardless of program environment or invocation.
```
prashantv, line 1523: I think deterministic makes a lot of sense, although I'm not sure about environment. E.g., parsing environment variables to set up some global?

often done for things like gRPC log level control via environment variable, or optional overrides to a default hostname / port based on env vars, etc.

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+2. Be completely self-contained within the program.
```
prashantv, line 1524: What do you mean by self-contained -- no dependencies?

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+3. Not depend on side-effects of other `init()` functions.
+4. Be able to run in any order. While `init()` ordering is well-known, code can
```
prashantv, line 1526: does (3) imply (4)? maybe they can be merged?

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+   change, and thus `init()` functions that rely on specific sequencing make
+   code brittle and error-prone.
+5. Not access or manipulate global or environment state, such as machine
+   information, environment variables, working directory, program
+   arguments/inputs, etc.
+6. Not perform I/O, including both filesystem, network, and system calls.
```
prashantv, line 1532: I think (1), (5) and (6) have a bit of overlap, so we may be able to merge them.

I can definitely see `init` functions being useful that violate (1), (5), and (6) (e.g., creating a variable parsed from an environment var, or parsing some data). Only really used when dealing with globals typically.

Since these requirements are marked as "At a minimum, all code [..] should", I think it might be a stronger message than the typical "avoid" messaging.

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+}
+```
+
+</td></tr>
+</tbody></table>
+
+Considering the above, some situations in which `init()` may be preferable or
+necessary might include:
+
+- Complex expressions that cannot be represented as single assignments.
```
prashantv, line 1619: one trick we might want an example for: have a global variable initialized using a function, e.g.,
```
var _Config = parseConfig()
```

that way, we avoid an `init`, and the function is testable

style.md
```diff
@@ -1515,6 +1516,114 @@ Note that the compiler will not generate errors when using predeclared
...
+- Seeding or constructing default helpers or state, such as `math/rand.Seed`.
```
prashantv, line 1620: This violates the "don't manipulate global state" recommendation. I agree it's useful to seed, but I think it sends a bit of a mixed message

style.md
mway: I don't think there's anything wrong with using environment variables, I think the question is whether or not we recommend avoiding environment variables _in `init()`_. Why do those things need to happen in `init()`? Why not as a default config-providing function, or as part of a ctor, or as part of config loading, etc?

In your example, the problem that we quickly run into is an (ostensible) dichotomy between "provide explicit configuration paths through exported types" and "provide implicit global configuration". This is complicated further by the fact that, if you do some "init-level config", you're probably bifurcating your user-facing config (like gRPC does, which maybe isn't a model to follow).

What do you think?

style.md
mway: Yeah, but I don't think this provides value over "no IO", et al. I'll remove it.

style.md
mway: I agree, but also think that if we merge them we end up with either (a) sparse guidance that doesn't really speak to all of the elements involved or (b) a mega-bullet. I'll see if there's a better way to combine them that condenses without losing meaning.

RE "all code should" vs "avoid", agreed. We should use more permissive language.

style.md
mway: Agreed, called this out above in the `_defaultFoo` example.

style.md
mway: Reworked these a bit, let me know what you think. I left state and IO as separate points, it felt a little dense when giving examples of both, and I think they're distinct enough that separating them doesn't feel too redundant.

style.md
prashantv: True, explicitly passing configuration is a lot cleaner. I think `init` is often used to configure globals or singletons which might not have an initializer.

That said, it's not a good pattern, so discouraging it makes sense to me.

PR #79, opened 2020-01-02
prefer specifying slice capacity for appending

hothero:

PR #97, opened 2020-06-13
Combine capacity guidance for maps and slices

mway:
Per @prashantv's [feedback](https://github.com/uber-go/guide/pull/79#pullrequestreview-380021958) on #79:

> Suggestion: Rather than having 2 separate topics with similar guidelines, what if this was one guideline with sub-headings for maps vs slices.
> 
> We can have the common reasons, then have type-specific examples + reasons in the sub-headings.

This also expands a bit more on map allocations at initialization time.

### DIFF
--- a/style.md
+++ b/style.md
@@ -76,7 +76,7 @@ row before the </tbody></table> line.
   - [Prefer strconv over fmt](#prefer-strconv-over-fmt)
   - [Avoid string-to-byte conversion](#avoid-string-to-byte-conversion)
   - [Prefer Specifying Container Capacity](#prefer-specifying-container-capacity)
-      - [Specifying Map Capacity](#specifying-map-capacity)
+      - [Specifying Map Capacity Hints](#specifying-map-capacity-hints)
       - [Specifying Slice Capacity](#specifying-slice-capacity)
 - [Style](#style)
   - [Be Consistent](#be-consistent)
@@ -1725,11 +1725,11 @@ BenchmarkGood-4  500000000   3.25 ns/op
 
 ### Prefer Specifying Container Capacity
 
-Specify container capacity where possible, as the compiler will allocate all
-required memory for that capacity up front rather than potentially allocating
-it in chunks as the container grows.
+Specify container capacity where possible in order to allocate memory for the
+container up front.  This minimizes subsequent allocations (by copying and
+resizing of the container) as elements are added.
 
-#### Specifying Map Capacity
+#### Specifying Map Capacity Hints
 
 Where possible, provide capacity hints when initializing
 maps with `make()`.
@@ -1743,12 +1743,9 @@ map at initialization time, which reduces the need for growing
 the map and allocations as elements are added to the map.
 
 Note that, unlike slices, map capacity hints do not guarantee complete,
-preemptive allocation: preallocation only occurs if the specified capacity
-indicates that additional hashmap buckets will need to be allocated internally.
-This means that map allocations may still occur when adding elements up to the
-specified capacity to the map. See [makemap()] for more information.
-
-  [makemap()]: https://golang.org/src/runtime/map.go
+preemptive allocation, but are used to approximate the number of hashmap
+buckets required. Consequently, means that map allocations may still occur when
+adding elements to the map, even up to the specified capacity.
 
 <table>
 <thead><tr><th>Bad</th><th>Good</th></tr></thead>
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -1719,7 +1720,13 @@ BenchmarkGood-4  500000000   3.25 ns/op
 </td></tr>
 </tbody></table>
 
-### Prefer Specifying Map Capacity Hints
+### Prefer Specifying Container Capacity
+
+Specify container capacity where possible, as the compiler will allocate all
+required memory for that capacity up front rather than potentially allocating
+it in chunks as the container grows.
```
prashantv, line 1727: The mention of the compiler seems a little too detailed here (since technically, most of the performance guidance affects what the compiler will do, though we don't call it out explicitly).

I also want to emphasize that it's not just that we allocate the memory in chunks (it might seem like instead of allocating 1mb, we'll allocate 5x200kb), but that we'll end up constantly resizing which increases total allocations, and also total time spent due to copies.

Perhaps:
"Specify container capacity where possible to allocate all required memory for the container up front. This avoids copies + resizing of the container as elements are added."

would replace mention of the compiler here (since it can happen at runtime for heap allocations), and the issue isn't just the chunks, but the resizing (e.g., overall more memory allocated + copies)

style.md
```diff
@@ -1719,7 +1720,13 @@ BenchmarkGood-4  500000000   3.25 ns/op
...
+
+#### Specifying Map Capacity
```
prashantv, line 1729: s/Capacity/Capacity Hint/

emphasizes that it's just a hint

style.md
```diff
@@ -1730,9 +1737,15 @@ make(map[T1]T2, hint)
...
-the map and allocations as elements are added to the map. Note
-that the capacity hint is not guaranteed for maps, so adding
-elements may still allocate even if a capacity hint is provided.
+the map and allocations as elements are added to the map.
+
+Note that, unlike slices, map capacity hints do not guarantee complete,
+preemptive allocation: preallocation only occurs if the specified capacity
+indicates that additional hashmap buckets will need to be allocated internally.
+This means that map allocations may still occur when adding elements up to the
+specified capacity to the map. See [makemap()] for more information.
```
prashantv, line 1746: Should we try and keep this a little more detached from the implementation details, maybe a simpler summary such as:
"map capacity hints do not guarantee complete, preemptive allocation, but are used to approximate the number of hashmap buckets required" or something along those lines?

style.md
```diff
@@ -1774,9 +1787,19 @@ allocations at assignment time.
...
+
+```go
+make([]T, length, hint)
+```
 
-Where possible, provide a capacity value to `make()` when initializing a slice for appending.
+Unlike maps, slice capacity is not a hint: the compiler will allocate enough
+memory for the capacity of the slice as provided to `make()`, which means that
+subsequent `append()` operations will incur zero allocations (until the capacity
+of the slice needs to be expanded to hold the appended element).
```
prashantv, line 1802: "needs to be expanded" may be clearer by using "length == capacity" condition

```suggestion
subsequent `append()` operations will incur zero allocations (until the length
of the slice matches the capacity, which will require a resize to hold additional
elements)
```

style.md
mway: Yeah, I think you're right. I generalized this a bit more to apply more correctly to both slices and maps (since map capacity hints don't guarantee that you'll "allocate all required memory for the container up front" obv) - not trying to be overly pedantic, just want the text to be accurate. Let me know what you think.

style.md
```diff
@@ -1722,7 +1723,13 @@ BenchmarkGood-4  500000000   3.25 ns/op
 </td></tr>
 </tbody></table>
 
-### Prefer Specifying Map Capacity Hints
+### Prefer Specifying Container Capacity
+
+Specify container capacity where possible in order to allocate memory for the
+container up front.  This minimizes subsequent allocations (by copying and
```
prashantv, line 1729: super nit: I think there's two spaces before "This"

style.md
```diff
@@ -1777,9 +1787,20 @@ allocations at assignment time.
...
+
+```go
+make([]T, length, capacity)
+```
 
-Where possible, provide a capacity value to `make()` when initializing a slice for appending.
+Unlike maps, slice capacity is not a hint: the compiler will allocate enough
+memory for the capacity of the slice as provided to `make()`, which means that
+subsequent `append()` operations will incur zero allocations (until the length
+of the slice matches the capacity, which will require a resize to hold
```
prashantv, line 1802: s/which will require a resize/after which any appends will require a resize/

PR #102, opened 2020-07-23
Guideline: Pointers to Interfaces

### LINKED ISSUE #101: Guideline: Pointers to Interfaces
If you want interface methods to modify the underlying data, you must use a **pointer**.

which pointer? pointer of object in assignment statement.

63isOK:
closes #101

### DIFF
--- a/style.md
+++ b/style.md
@@ -146,7 +146,26 @@ An interface is two fields:
   the data stored is a value, then a pointer to the value is stored.
 
 If you want interface methods to modify the underlying data, you must use a
-pointer.
+pointer of object in assignment statement.
+
+```go
+type F interface {
+  f()
+}
+
+type S1 struct{}
+
+func (s S1) f() {}
+
+type S2 struct{}
+
+func (s *S2) f() {}
+
+// f1.f() can't modify the underlying data
+// f2.f() can modify the underlying data
+var f1 F:= S1{}
+var f2 F:= &S2{}
+```
 
 ### Receivers and Interfaces
 


style.md
```diff
@@ -154,7 +154,26 @@ An interface is two fields:
   the data stored is a value, then a pointer to the value is stored.
 
 If you want interface methods to modify the underlying data, you must use a
-pointer.
+pointer of object in assignment statement.
```
mway, line 157: This isn't exactly true: if the method receiver is a pointer but the type being used is an _addressable value_, it will be implicitly converted to a pointer for that call:

```go
type Foo struct {
  bar string
}

func (f *Foo) SetBar(value string) {
  f.bar = value
}

func (f *Foo) Bar() string {
  return f.bar
}

var f Foo
f.SetBar("hello")

fmt.Println(f.Bar())
// prints "hello"
```

You can see an example [here](https://play.golang.org/p/EX3EvSU0D2e).

TudorHulban: The example with property string and pointer maybe it could be changed as why would one use a pointer receiver on a primitive type?

mway: Choosing when to use a pointer or a value receiver is different than satisfying interfaces with pointer or value types (just so we don't conflate that with this PR).

That said, the official Go convention for using pointer or value receivers is pretty unambiguous: see "[Should I define methods on values or pointers?](https://golang.org/doc/faq#methods_on_values_or_pointers)" in the [Go FAQ](https://golang.org/doc/faq).

mway:
Going to close this based on the earlier comments. Feel free to open an issue to discuss other potential clarifications!

PR #104, opened 2020-07-24
long name of package

### LINKED ISSUE #103: 'Import Aliasing'  is good for long name of package

63isOK:
closes #103

### DIFF
--- a/style.md
+++ b/style.md
@@ -2083,6 +2083,14 @@ import (
 )
 ```
 
+It is good for long name of package in some cases.
+
+```go
+import (
+  jww "github.com/spf13/jwalterweatherman"
+)
+```
+
 In all other scenarios, import aliases should be avoided unless there is a
 direct conflict between imports.
 


mway:
Hey @63isOK, thanks for the PR!

Import aliases are generally used in cases in which the package name doesn't reflect the last fragment of the import path - we do **not** recommend abbreviating package names arbitrarily, especially in potentially confusing ways. Aliases should be clarifying, not obfuscating, and the common Go convention is that package names match the last part of the import path.

If the length of a package name is a concern, folks may want to consider choosing terser package names in line with Go's [package naming guidelines](https://golang.org/doc/effective_go.html#package-names); if that's not possible, import aliases are always an option - we just don't offer guidance around this (and there probably isn't a much value in doing so unless this becomes a frequent issue or common point of confusion), as it's really a case-by-case and codebase-by-codebase sort of thing.

PR #105, opened 2020-09-14
Don't Panic: Remove main panic

abhinav:
In our example, we `panic` in `main` instead of the function `foo`. The
point that this was demonstrating was that the caller of the function
has the choice of panicking rather than `foo` itself.

Unfortunately, this causes confusion for readers modeling their `main()`
after this code, because the "good" example still panics.

This reduces confusion here by having `main` handle the `error` with an
`os.Exit`.

Note that this moves argument handling into `foo` because placing it in
`main` is not very testable.

### DIFF
--- a/style.md
+++ b/style.md
@@ -1077,40 +1077,33 @@ allow the caller to decide how to handle it.
 <tr><td>
 
 ```go
-func foo(bar string) {
-  if len(bar) == 0 {
-    panic("bar must not be empty")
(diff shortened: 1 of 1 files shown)


twilly:
Do you know where your towel is?

(+42)

mway (approved):

prashantv (approved):

PR #106, opened 2020-12-02
don't use interface pointer

xxjwxc:
don't use interface pointer


abhinav (approved):
Thanks! And a good catch on the returns.

I made a minor change: the first example is specifically for struct pointer embedding so I switched that back.

prashantv (approved):
LGTM, thanks for the fixes

PR #107, opened 2020-12-08
Add export option interface example

LiangXianSen:


abhinav:
Thanks for the proposal @LiangXianSen. I'm sorry but we've decided not to
accept this suggestion.

This is clever in allowing unifying all option types into one. However, we think that
this pattern reduces readability and loses the type safety guaranteed by the currently
recommend approach. We don't see a strong advantage of unifying the option type to
justify those costs.

LiangXianSen:
Sure, there is no strong advantage for program performance, but it shares a good program design for organizing your project. It shows a clear correlation between those packages. 

It from a real case when I am going to design a project, I met a problem that declare a interface on a package that may requires some optional parameters which are not fixed. I have not found any good pattern achieve this.  I saw option function and option interface are only suit for one specific package with fixed internal struct. Derived by option interface I made a litter different implements what I needs. Btw, I just shared with that. this pattern is more flexible.

```
├── option.go
├── sender
│   ├── dingding
│   │   ├── dingding.go
│   │   ├── dingding_test.go
│   │   ├── example_test.go
│   │   └── option.go
│   ├── email
│   │   ├── auth.go
│   │   ├── email.go
│   │   ├── email_test.go
│   │   ├── example_test.go
│   │   └── option.go
│   ├── sender.go
│   └── wechat
│       ├── example_test.go
│       ├── message.go
│       ├── option.go
│       └── wechat.go
```

PR #113, opened 2021-03-04
Additional guidance on struct initialization

### LINKED ISSUE #109: Initialization of fields with zero values
For struct initialization where some of the fields are zero values, we should establish whether we prefer for those fields to be initialized to zero values explicitly or omitted:

    foo := Foo{
      Count:   1,
      Items:   nil,
      Default: "",
      Admin:   true,
    }
    
    // or
    
    foo := Foo{
      Count: 1,
      Admin: true,
    }

Opinion: the latter is preferable because it reduces code noise from things that don't matter at this time. You only specify the things that have meaning/value and everything else is the default.

---

Follow up: When all fields are zero value, do we want to use the explicit struct initialization form or `var`?

    foo := Foo{}
    // or
    var foo Foo

Opinion: the latter because it scans differently (visually) and differentiates from `Foo{}` where `Foo` is a map or slice type (see also, [Initializing Maps][1]). It makes it clear that this is something that is probably not going to be used as-is and will be filled separately. (`{}` form to initialize structs still applies when not stored in a variable or fields are non-zero: `do(Stuff{...})`.)

[1]: https://github.com/uber-go/guide/blob/master/style.md#initializing-maps

abhinav:
This adds a new Struct Initialization section, moving the old "use field
names" and "initializing struct references" guidelines under it.

In the new section, based on #109, this adds the following new sections:

- Omit zero value fields
- Use var for zero value structs

Each commit is individually reviewable.

Preview: https://github.com/uber-go/guide/blob/abg/struct-init/style.md#initializing-structs
Resolves #109

### DIFF
--- a/style.md
+++ b/style.md
@@ -98,6 +98,7 @@ row before the </tbody></table> line.
   - [Use Raw String Literals to Avoid Escaping](#use-raw-string-literals-to-avoid-escaping)
   - [Initializing Structs](#initializing-structs)
       - [Use Field Names to Initialize Structs](#use-field-names-to-initialize-structs)
+      - [Omit Zero Value Fields in Structs](#omit-zero-value-fields-in-structs)
       - [Initializing Struct References](#initializing-struct-references)
   - [Initializing Maps](#initializing-maps)
   - [Format Strings outside Printf](#format-strings-outside-printf)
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -2837,6 +2838,54 @@ tests := []struct{
 }
 ```
 
+#### Omit Zero Value Fields in Structs
+
+When initializing structs with field names, omit fields that have zero values.
```
prashantv, line 2844: rather than calling out the exception, can we incorporate it into the recommendation, "omit fields that have zero values where the field doesn't provide meaningful context" or something like that?

prashantv (approved):
LGTM with small suggestion.

style.md
mway: Agree, despite calling it out below:

> This helps reduce noise for readers by omitting values that are default in
that context. Only meaningful values are specified.

```suggestion
When initializing structs with field names, omit fields that have zero values unless they provide meaningful context.
```

mway (approved):

style.md
abhinav: Good point. Addressed.

PR #114, opened 2021-03-04
Add guidance on exiting

### LINKED ISSUE #66: Guideline: One log.Fatal per application
This is to track adding a guideline around use of log.Fatal. Our
recommendation in code reviews has been to have exactly one log.Fatal per
application: in the main. Everything else should return an error.

abhinav:
In #66, we discussed advising at most one `log.Fatal` per program. This
can be split into two pieces of guidance:

1. exit only in main
2. exit only once

Applying to both, `os.Exit` and `log.Fatal`.

This adds (1) as a guidance to the style guide, and (2) as an optional
"upgrade" for extra points.

At minimum, (1) gives us easy-to-follow control flow and testability
outside main. If readers are sold on that, (2) gives us testable
`main()` logic.

Preview: https://github.com/uber-go/guide/blob/abg/exit-once/style.md#exit-in-main
Resolves #66

### DIFF
--- a/style.md
+++ b/style.md
@@ -72,6 +72,7 @@ row before the </tbody></table> line.
   - [Avoid Embedding Types in Public Structs](#avoid-embedding-types-in-public-structs)
   - [Avoid Using Built-In Names](#avoid-using-built-in-names)
   - [Avoid `init()`](#avoid-init)
+  - [Exit in Main](#exit-in-main)
 - [Performance](#performance)
   - [Prefer strconv over fmt](#prefer-strconv-over-fmt)
   - [Avoid string-to-byte conversion](#avoid-string-to-byte-conversion)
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -1632,6 +1633,98 @@ necessary might include:
 
   [Google Cloud Functions]: https://cloud.google.com/functions/docs/bestpractices/tips#use_global_variables_to_reuse_objects_in_future_invocations
 
+### Exit Once in Main
+
+Go programs use [`os.Exit`] or [`log.Fatal*`] to exit. (Panicking is not a
```
prashantv, line 1638: or return from `main`, perhaps "exit immediately"?

prashantv (commented):
+1 on the general direction of this guidance, but also want to call out a common reason why multiple exits can be useful. With structured logging, a `logger.Fatal` may have additional fields with context for something failed, and returning it as an `error` means the structured information is flatted into a string.

This is also a similar problem resulting in code logging *and* returning an error (which we also discourage).

abhinav:
Updated based on discussion. Guidance split into two:

1. exit only in main
2. optionally, if possible, exit only once

The two pieces are in separate commits if we decide we only want one of
these for now.

prashantv (approved):

PR #117, opened 2021-03-10
Go Common Mistakes: Title case

abhinav:
Title case the link for Go Common Mistakes guide to match format of
other links.


prashantv (approved):
Thanks!

PR #119, opened 2021-03-15
refactor: Added errors is

alexander-melentyev:
Hello!
I suggest you use errors.Is in good example


abhinav (approved):
Reasonable. Thanks!

prashantv (approved):

PR #123, opened 2021-05-15
refactor: change golint to revive

EpicStep:
golangci-lint recommend change golint to revive

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,7 +9,7 @@ linters:
   enable:
     - errcheck
     - goimports
-    - golint
+    - revive
     - govet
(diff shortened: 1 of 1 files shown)


mway:
Similar comment to #124:

We haven't evaluated revive internally yet (to my knowledge, anyway), though on paper it does look promising.

We don't recommend things that we don't do/use ourselves, so I'd prefer to wait on mentioning/recommending revive until we've had a chance to look into/adopt it internally first. I'll start a conversation internally about this, and we can leave the PR open in the meantime. (Note that whenever we do make this change, it should also include #124.)

PR #124, opened 2021-05-15
Refactor: change golint to revive in styleguide.md

gvencadze:
golint is deprecated and recommends switch to [revive](https://github.com/mgechev/revive)

### DIFF
--- a/style.md
+++ b/style.md
@@ -137,11 +137,11 @@ resources:
 2. [Go Common Mistakes](https://github.com/golang/go/wiki/CommonMistakes)
 3. [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments)
 
-All code should be error-free when run through `golint` and `go vet`. We
+All code should be error-free when run through `revive` and `go vet`. We
 recommend setting up your editor to:
 
 - Run `goimports` on save
-- Run `golint` and `go vet` to check for errors
+- Run `revive` and `go vet` to check for errors
 
 You can find information in editor support for Go tools here:
 <https://github.com/golang/go/wiki/IDEsAndTextEditorPlugins>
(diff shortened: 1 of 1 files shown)


prashantv:
I wasn't able to find any recommendations on the [golint README](https://github.com/golang/lint) to switch to revive, can you point to where this is recommended? The only recommendation I was able to find is:

> NOTE: Golint is deprecated and frozen. There's no drop-in replacement for it, but tools such as Staticcheck and go vet should be used instead.

gvencadze:
Recommendation was showed in log of running golint in github CI. Or better I should just add link to revive (instead of removing) and write that golint is deprecated?

> level=warning msg="[runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive."

mway:
We haven't evaluated revive internally yet (to my knowledge, anyway), though on paper it does look promising.

We don't recommend things that we don't do/use ourselves, so I'd prefer to wait on mentioning/recommending revive until we've had a chance to look into/adopt it internally first. I'll start a conversation internally about this, and we can leave the PR open in the meantime. (Note that whenever we do make this change, it should also include #123.)

conrad760:
lurking, was a consensus ever drawn from this? 
How do you enforce your style guides internally?

abhinav:
FWIW, I've switched from golint to revive in some of my personal projects.
It's almost a drop-in replacement minus:

- it does not exit with a non-zero status code by default if there are issues
- the default rule set as of revive 1.3.0 does a bit more than golint, and we might not agree with all of those rules

If this style guide recommends revive, we should provide a sample recommended configuration.

PR #130, opened 2021-07-08
Mutexes: Advise against embedding

### LINKED ISSUE #127: Embedding a sync.Mutex is equally bad whether or not the type is exported
> Unexported structs that use a mutex to protect fields of the struct may embed the mutex.

What makes unexported types special? 🙃 

A mutex in a struct is almost always an implementation detail, used to provide concurrency-safety to the capabilities (methods) of the type. Embedding it allows callers to manipulate it directly, but that's an abstraction leak, breaks encapsulation, and is just an enormous maintenance risk, for what I hope are obvious reasons. Whatever callers are doing when they take and release the lock can equally well be expressed as a method, right?

Is there some other use case you have that motivates this exception?

abhinav:
Our guidance for mutexes suggested that it was okay to embed mutexes in
structs if the struct was unexported. This was left as an allowance for
"if you really want to do it, this is the only time it's okay", but that
nuance doesn't translate well to the style guide.

Advise against embedding mutexes completely to resolve any confusion
here.

Resolves #127

### DIFF
--- a/style.md
+++ b/style.md
@@ -333,7 +333,7 @@ mu.Lock()
 </tbody></table>
 
 If you use a struct by pointer, then the mutex should be a non-pointer field on
-it. Do not embed the mutex on the struct.
+it. Do not embed the mutex on the struct, even if the struct is not exported.
 
 <table>
 <thead><tr><th>Bad</th><th>Good</th></tr></thead>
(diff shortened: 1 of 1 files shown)


abhinav:
image

prashantv (approved):
Should we explicitly call out that "Even for unexported types, <reason>"?

PR #134, opened 2021-11-16
errors: Update for errors.Is/As

### LINKED ISSUE #132: Some guidelines in "Error Types" and "Error Wrapping" are incorrect
> - Wrapped errors using "pkg/errors".Wrap

Since Go 1.13, wrapping errors is canonically done with fmt.Errorf and %w.

Relatedly, in the Error wrapping section,

```go
return fmt.Errorf("new store: %v", err)
```

should probably use %w instead of %v — or, if %v was intentional, should describe when to use %v vs. %w.

—

> Do the clients need to detect and handle this error? If so, you should use a custom type, and implement the Error() method.

Custom types are appropriate only when consumers need extra information about the error beyond it's identity. If consumers only need to handle the error by identity, then the more appropriate choice is a sentinel error, i.e. `var ErrFoo = errors.New("...")`. [ref](http://peter.bourgon.org/blog/2019/09/11/programming-with-errors.html)

—

> Be careful with exporting custom error types directly since they become part of the public API of the package. It is preferable to expose matcher functions to check the error instead.

It's true that both exported sentinel errors and error types become part of your package API, but the same is true of the matcher function suggested here, and they both have ~equivalent impact on API compatibility. What's more, the consequent advice to return unexported error types e.g.

```go
type errNotFound struct {
  file string
}
```

is surely incorrect, because the extra information captured in the type is inaccessible to callers. (In fact, it's almost always an error to return an unexported type from an exported function or method.)

abhinav:
This PR updates the guidance we have for errors to reflect best practices since
Go 1.13's error APIs. The high-level overview of the guidance is:

- use `errors.New` for static strings,
  exporting a `var` if it needs to be matched
- declare a custom error type if it needs to be matched
- use `fmt.Errorf` for dynamic strings if it does not need to be matched
- use `fmt.Errorf` with `%w` when wrapping if the wrapped error needs to be
  matched
- name top-level error variables `errFoo` or `ErrFoo`
- name error types `FooError`

Resolves #132, #8
Ref GO-1014

---

[Preview](https://github.com/uber-go/guide/blob/error-wrapping/style.md)

### DIFF
--- a/style.md
+++ b/style.md
@@ -956,15 +956,15 @@ There are three main options for propagating errors if a call fails:
 
 - Return the original error if there is no additional context to add and you
   want to maintain the original error type.
-- Add context using [`"pkg/errors".Wrap`] so that the error message provides
-  more context and [`"pkg/errors".Cause`] can be used to extract the original
-  error.
-- Use [`fmt.Errorf`] if the callers do not need to detect or handle that
-  specific error case.
+- Add context using `fmt.Errorf` with a `%w` verb if you want callers to
+  be able to match against or extract the original error.
+  If you do this, note that this is now part of your public API.
+- Add context with `fmt.Errorf` with a `%v` verb if callers should not match
+  that error separately.
 
-It is recommended to add context where possible so that instead of a vague
-error such as "connection refused", you get more useful errors such as
-"call service foo: connection refused".
+We recommend adding context where possible so that
+instead of a vague error such as "connection refused",
+you get more useful errors such as "call service foo: connection refused".
 
 When adding context to returned errors, keep the context succinct by avoiding
 phrases like "failed to", which state the obvious and pile up as the error
@@ -979,7 +979,7 @@ percolates up through the stack:
 s, err := store.New()
 if err != nil {
     return fmt.Errorf(
-        "failed to create new store: %v", err)
+        "failed to create new store: %w", err)
 }
 ```
 
@@ -989,7 +989,7 @@ if err != nil {
 s, err := store.New()
 if err != nil {
     return fmt.Errorf(
-        "new store: %v", err)
+        "new store: %w", err)
 }
 ```
 
--- a/style.md
+++ b/style.md
@@ -950,8 +950,8 @@ if err := foo.Open("testfile.txt"); err != nil {
 </td></tr>
 </tbody></table>
 
-Be careful with exporting error variables or types.
-These will become part of the public API of your package.
+Note that if you export error variables or types from a package,
+they will become part of the public API of the package.
 
 #### Error Wrapping
 
@@ -1027,21 +1027,48 @@ This guidance supersedes the [Prefix Unexported Globals with _](#prefix-unexport
 
 ```go
 var (
-  errNotFound = errors.New("not found")
+  // The following two errors are exported
+  // so that users of this package can match them
+  // with errors.Is.
+
   ErrBrokenLink = errors.New("link is broken")
   ErrCouldNotOpen = errors.New("could not open")
+
+  // This error is not exported because
+  // we don't want to make it part of our public API.
+  // We may still use it inside the package
+  // with errors.Is.
+
+  errNotFound = errors.New("not found")
 )
 ```
 
 For custom error types, use the suffix `Error` instead.
 
 ```go
+// Similarly, this error is exported
+// so that users of this package can match it
+// with errors.As.
+
 type NotFoundError struct {
   File string
 }
 
 func (e NotFoundError) Error() string {
   return fmt.Sprintf("file %q not found", e.File)
+}
+
+// And this error is not exported because
+// we don't want to make it part of the public API.
+// We can still use it inside the package
+// with errors.As.
+
+type resolveError struct {
+  Path string
+}
+
+func (e resolveError) Error() string {
+  return fmt.Sprintf("resolve %q", e.Path)
 }
 ```
 
--- a/style.md
+++ b/style.md
@@ -822,7 +822,7 @@ Consider the following before picking the option best suited for your use case.
 
 - Does the caller need to match the error so that they can handle it?
   If yes, we must support the [`errors.Is`] and [`errors.As`] functions
-  by exporting an error variable or type.
+  by declaring a top-level error variable or a custom type.
 - Is the error message a static string,
   or is it a dynamic string that requires contextual information?
   For the former, we can use [`errors.New`], but for the latter we must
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -811,31 +813,40 @@ seconds that may have occurred between those two instants.
...
+[`errors.As`]: https://golang.org/pkg/errors/#As
 
-  [`errors.New`]: https://golang.org/pkg/errors/#New
-  [`fmt.Errorf`]: https://golang.org/pkg/fmt/#Errorf
-  [`"pkg/errors".Wrap`]: https://godoc.org/github.com/pkg/errors#Wrap
+| Error matching? | Error Message | Guidance                     |
+|-----------------|---------------|------------------------------|
+| No              | static        | [`errors.New`]               |
+| No              | dynamic       | [`fmt.Errorf`]               |
+| Yes             | static        | variable with [`errors.New`] |
```
rabbbit, line 840: what do you mean by `variable` here? Did you mean `exported variable`?

In the example below you had
```
var (
  errNotFound = errors.New("not found")
  ErrBrokenLink = errors.New("link is broken")
  ErrCouldNotOpen = errors.New("could not open")
)
```
so all of them are variables?

abhinav: Yeah, they're all variables. Exported or not depends on whether you want to export or not. I'll change the guidance there to say `var` and document the examples further.

rabbbit: Yes. I wasn't sure why just one of them calls out `variable`, but others don't.

rabbbit: shouldn't the option (1) also be a global variable, just not exported? Avoids an allocation AFAIU ?

abhinav: Depends on re-use. If you're never matching that error, it's a bit annoying and noisy to put every error string into an errors.New.

RE: allocation: Yeah, putting it in a global would save an alloc, but is the inconvenience of pulling every error message out like that worth the savings? I suspect performance sensitive contexts will do that as needed when optimizing their allocations.

style.md
```diff
@@ -63,8 +63,10 @@ row before the </tbody></table> line.
   - [Channel Size is One or None](#channel-size-is-one-or-none)
   - [Start Enums at One](#start-enums-at-one)
   - [Use `"time"` to handle time](#use-time-to-handle-time)
-  - [Error Types](#error-types)
-  - [Error Wrapping](#error-wrapping)
+  - [Errors](#errors)
+    - [Error Types](#error-types)
+    - [Error Wrapping](#error-wrapping)
+    - [Error naming](#error-naming)
```
mway, line 69: nit: consistent Title Case

style.md
```diff
@@ -811,34 +813,43 @@ seconds that may have occurred between those two instants.
...
-When returning errors, consider the following to determine the best choice:
+- Does the caller need to match the error so that they can handle it?
+  If yes, we must support the [`errors.Is`] and [`errors.As`] functions
+  by declaring a top-level error variable or a custom type.
+- Is the error message a static string,
+  or is it a dynamic string that requires contextual information?
+  For the former, we can use [`errors.New`], but for the latter we must
+  use [`fmt.Errorf`] or a custom error type.
+- Are we propagating a new error returned by a downstream function?
+  If so, see the [section on error wrapping](#error-wrapping).
```
mway, line 831: semi-meta, but the perspective here is first person; ideally it's consistent throughout the guide (it's not atm). we should think about what perspective makes sense/is the most accessible to read.

style.md
```diff
@@ -1044,6 +1019,59 @@ See also [Don't just check errors, handle them gracefully].
   [`"pkg/errors".Cause`]: https://godoc.org/github.com/pkg/errors#Cause
   [Don't just check errors, handle them gracefully]: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully
 
+#### Error naming
```
mway, line 1022: nit: Title Case

mway (approved):
lgtm with minor nit

nomis52 (approved):

style.md
abhinav: Yeah. It's been a mix of "do this" and "we should do this" and "this should be done".

style.md
```diff
@@ -811,34 +813,43 @@ seconds that may have occurred between those two instants.
...
-- [`errors.New`] for errors with simple static strings
-- [`fmt.Errorf`] for formatted error strings
-- Custom types that implement an `Error()` method
-- Wrapped errors using [`"pkg/errors".Wrap`]
+There are few options for declaring errors.
+Consider the following before picking the option best suited for your use case.
 
-When returning errors, consider the following to determine the best choice:
+- Does the caller need to match the error so that they can handle it?
+  If yes, we must support the [`errors.Is`] and [`errors.As`] functions
```
peterbourgon, line 824: s/and/or/ ?

style.md
```diff
@@ -887,112 +893,81 @@ if err := foo.Open(); err != nil {
...
-func (e errNotFound) Error() string {
-  return fmt.Sprintf("file %q not found", e.file)
+type NotFoundError struct {
+  File string
 }
 
-func IsNotFoundError(err error) bool {
-  _, ok := err.(errNotFound)
-  return ok
+func (e NotFoundError) Error() string {
```
peterbourgon, line 929: Methods to satisfy the error interface should always be [defined on the pointer receiver](https://stackoverflow.com/questions/50333428/custom-errors-in-golang-and-pointer-receivers).

style.md
```diff
@@ -887,112 +893,81 @@ if err := foo.Open(); err != nil {
...
 - Return the original error if there is no additional context to add and you
   want to maintain the original error type.
-- Add context using [`"pkg/errors".Wrap`] so that the error message provides
-  more context and [`"pkg/errors".Cause`] can be used to extract the original
-  error.
-- Use [`fmt.Errorf`] if the callers do not need to detect or handle that
-  specific error case.
+- Add context using `fmt.Errorf` with a `%w` verb if you want callers to
+  be able to match against or extract the original error.
+  If you do this, note that this is now part of your public API.
```
peterbourgon, line 964: Not unless you say it is :)

style.md
abhinav: Good call! Fixing.

style.md
abhinav: I disagree on that as an "always" rule. It's up to you, based on what semantics you want.
That said, I agree that "if unsure, use pointer receiver" is a good rule of thumb.
Since it's difficult to communicate that nuance in the style guide,
we can change the example code to use a pointer receiver without making an explicit statement about it.

style.md
abhinav: Yes, but I mean, Hyrum's Law applies. If you were exposing it before, and users rely on having the underlying cause available, changing that is a breaking change regardless of whether you documented it. But I see your point. I'll see if there's a way to reword this a little so authors are deliberate about the choice of using %v vs %w.

peterbourgon: Hyrum's Law cannot dictate API compatibility. If it did, then literally any user-visible change is a breaking change, which is facially incoherent.

style.md
peterbourgon: It's unfortunately more subtle than that -- there are specific technical reasons why error methods defined on value receivers introduce ambiguities that break callers, though I'm having trouble pulling up a good reference...

style.md
abhinav: Fair point. I guess I'm trying to find a way to suggest that the choice should be deliberate. I'm trying to work this section up a bit more.

peterbourgon: IME it's quite difficult to have enough knowledge/context to be able to make a deliberate choice on stuff like this! 

$0.02 — I've always suggested to use `%w` by default, and `%v` only when you explicitly want to hide information.

abhinav: Also reasonable. Internal discussion about this one yielded a similar result.
I think I've found a good balance of "use `%w` but be deliberate about it when necessary".
PR incoming.

abhinav: 134

abhinav: https://github.com/uber-go/guide/pull/135

PR #136, opened 2021-11-21
Fix table tags

ernado:
Close `<td>` or `<tr>`, this should help with `mdx` compatibility.
Also add missed newline.


mway:
Thank you!

PR #137, opened 2021-12-03
Clarify mutexes should not be embedded

rabbbit:
The instructions were contradictory:
- description say "embedded structs such as mutexes"
- later on, examples lists mutexes as something that shouldn't be embedded.

Let's clarify.

### DIFF
--- a/style.md
+++ b/style.md
@@ -509,7 +509,8 @@ snapshot := stats.Snapshot()
 
 ### Defer to Clean Up
 
-Use defer to clean up resources such as files and locks.
+Use defer to clean up resources 
+files and locks.
 
 <table>
 <thead><tr><th>Bad</th><th>Good</th></tr></thead>
@@ -2561,7 +2562,7 @@ See [Error Naming](#error-naming).
 
 ### Embedding in Structs
 
-Embedded types (such as mutexes) should be at the top of the field list of a
+Embedded types should be at the top of the field list of a
 struct, and there must be an empty line separating embedded fields from regular
 fields.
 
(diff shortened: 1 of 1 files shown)


style.md
```diff
@@ -2593,6 +2593,7 @@ type Client struct {
 Embedding should provide tangible benefit, like adding or augmenting
 functionality in a semantically-appropriate way. It should do this with zero
 adverse user-facing effects (see also: [Avoid Embedding Types in Public Structs]).
+Generally, mutexes should not be embedded.
```
rabbbit, line 2596: I'm also okay with removing this line - reading the description provides the same information.

However, given the prevalence of embedded mutexes, I thought it might be worth calling it out explicitly.

mway, line 2596:
```suggestion
Generally, mutexes should not be embedded unless they are an integral part of the embedding type's API.
```

mway: We previously called out mutexes above, so moving it here instead is fine - we can always remove it later if we decide it's not adding any value.

rabbbit:
cc @abhinav - landing, but pretty sure you'll agree wit this.

mway (approved):

style.md
```diff
@@ -2593,6 +2593,7 @@ type Client struct {
 Embedding should provide tangible benefit, like adding or augmenting
 functionality in a semantically-appropriate way. It should do this with zero
 adverse user-facing effects (see also: [Avoid Embedding Types in Public Structs]).
+Generally, mutexes should not be embedded unless they are an integral part of the embedding type's API.
```
abhinav, line 2596: They almost never are.
Let's change this to:

```suggestion
Exception: Mutexes should not be embedded even on unexported types. See also [Zero-value Mutexes are Valid](#zero-value-mutexes-are-valid).
```

PR #150, opened 2022-05-06
Fix typo

shirakiyo:
Remove repeated sentence.


sywhang (approved):
Wasn't clear from the commit message where the repeated sentence is, but I see it at L2646. Thanks for the PR.

PR #158, opened 2022-10-17
Add guidance on goroutine lifecycle management

abhinav:
This is mostly an RFC based on some internal discussions.
In general, we prefer for goroutines to have well-managed lifecycles.
No uncontrolled background work that cannot be stopped.

This change tries to distill some of the guidance around it into a style
guide entry.
It's not super comprehensive, but it tries to cover some of the guiding
principles.

Rendered: https://github.com/uber-go/guide/blob/goroutines/style.md#dont-fire-and-forget-goroutines

### DIFF
--- a/style.md
+++ b/style.md
@@ -78,6 +78,7 @@ row before the </tbody></table> line.
     - [Exit Once](#exit-once)
   - [Use field tags in marshaled structs](#use-field-tags-in-marshaled-structs)
   - [Don't fire-and-forget goroutines](#dont-fire-and-forget-goroutines)
+    - [Wait for goroutines to exit](#wait-for-goroutines-to-exit)
     - [No goroutines in `init()`](#no-goroutines-in-init)
 - [Performance](#performance)
   - [Prefer strconv over fmt](#prefer-strconv-over-fmt)
@@ -1974,8 +1975,9 @@ and we can wait for it to exit with `<-done`.
 </td></tr>
 </tbody></table>
 
-Given a goroutine with a predictable lifetime,
-or a goroutine that can be stopped with a signal,
+#### Wait for goroutines to exit
+
+Given a goroutine spawned by the sytem,
 there must be a way to wait for the goroutine to exit.
 There are two popular ways to do this:
 
--- a/style.md
+++ b/style.md
@@ -1898,8 +1898,8 @@ renaming fields.
 
 ### Don't fire-and-forget goroutines
 
-Goroutines are lightweight, but they're not free.
-Do not leak goroutines in production code.
+Goroutines are lightweight, but they're not free. They cost memory.
+Leaking goroutines slowly consume more and more memory.
 In general, every goroutine:
 
 - must have a predictable time at which it will stop running; or
--- a/style.md
+++ b/style.md
@@ -78,7 +78,7 @@ row before the </tbody></table> line.
     - [Exit Once](#exit-once)
   - [Use field tags in marshaled structs](#use-field-tags-in-marshaled-structs)
   - [Don't fire-and-forget goroutines](#dont-fire-and-forget-goroutines)
-    - [No goroutines in `init()`](#no-goroutines-in-init )
+    - [No goroutines in `init()`](#no-goroutines-in-init)
 - [Performance](#performance)
   - [Prefer strconv over fmt](#prefer-strconv-over-fmt)
   - [Avoid string-to-byte conversion](#avoid-string-to-byte-conversion)
--- a/style.md
+++ b/style.md
@@ -2046,13 +2046,20 @@ func doWork() {
 type Worker struct{ /* ... */ }
 
 func NewWorker(...) *Worker {
-  w := &Worker{/* .. */}
+  w := &Worker{
+    stop: make(chan struct{}),
+    done: make(chan struct{}),
+    // ...
+  }
   go w.doWork()
 }
 
 func (w *Worker) doWork() {
+  defer close(w.done)
   for {
     // ...
+    case <-w.stop:
+      return
   }
 }
 
@@ -2067,7 +2074,7 @@ func (w *Worker) Shutdown() {
 </td></tr>
 <tr><td>
 
-Spawns a background goroutine unconditionally if the user exports this package.
+Spawns a background goroutine unconditionally when the user exports this package.
 The user has no control over the goroutine or a means of stopping it.
 
 </td><td>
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -1894,6 +1896,167 @@ this contract. Specifying field names inside tags makes the contract explicit,
...
+      // ...
+    }()
+    
+    // To wait for the goroutine to finish:
+    <-done
+    ```
+
+#### No goroutines in `init()`
+
+`init()` functions should almost never spawn goroutines.
```
alxn, line 1995: it says "almost never", but doesn't give an example of when it may be possible (within an `init()` call)

style.md
```diff
@@ -1894,6 +1896,167 @@ this contract. Specifying field names inside tags makes the contract explicit,
 and it guards against accidentally breaking the contract by refactoring or
 renaming fields.
 
+### Don't fire-and-forget goroutines
+
+Goroutines are lightweight, but they're not free.
+Do not leak goroutines in production code.
```
alxn, line 1902: can we get a reference to https://github.com/uber-go/goleak ?

style.md
abhinav: I'll drop the "almost never" until we can come up with reasonable exceptions.

style.md
```diff
@@ -1894,6 +1896,172 @@ this contract. Specifying field names inside tags makes the contract explicit,
 and it guards against accidentally breaking the contract by refactoring or
 renaming fields.
 
+### Don't fire-and-forget goroutines
+
+Goroutines are lightweight, but they're not free. They cost memory.
+Leaking goroutines slowly consume more and more memory.
```
prashantv, line 1902: This makes it seem like it's the goroutine's memory usage that is the only concern, perhaps worth calling out that it's also any resources that may be referenced (E.g., timers, files, etc).

style.md
```diff
@@ -1894,6 +1896,172 @@ this contract. Specifying field names inside tags makes the contract explicit,
...
+  for {
+    select {
+    case <-tick.C:
+      flush()
+    case <-stop:
+      return
+    }
+  }
+}()
+```
```
prashantv, line 1948: what do you think of expanding the example a little to show a `Stop` method as well? I often see implementations where there's a `close(stop)`, but I usually recommend that the stop method waits for the background goroutine to complete.

style.md
```diff
@@ -1894,6 +1896,172 @@ this contract. Specifying field names inside tags makes the contract explicit,
...
+
+</td><td>
+
+This goroutine can be stopped with `close(stop)`.
+
+</td></tr>
+</tbody></table>
+
+Given a means of stopping the goroutine,
+there must be a way to wait for it to stop.
```
prashantv, line 1964: stopping the goroutine --> wait for it to stop is a little confusing to read, perhaps "wait for it to complete" or "exit"?

prashantv (approved):
Thanks for adding this, I've been thinking about this same recommendation after leaving multiple reviews in a similar vein!

style.md
mway: Agree, maybe something to the effect of:

> Goroutines are lightweight, but they're not free: at a minimum, they cost memory (for their stack) and CPU (due to scheduling). While these costs are typically negligible, goroutines can cause performance issues when spawned in large numbers or without controlling their lifetimes. Additionally, because goroutines can hold pointers that prevent GC for objects visible to them, unmanaged goroutine lifetimes can lead to memory leaks.

style.md
```diff
@@ -1894,6 +1896,172 @@ this contract. Specifying field names inside tags makes the contract explicit,
 and it guards against accidentally breaking the contract by refactoring or
 renaming fields.
 
+### Don't fire-and-forget goroutines
```
mway, line 1899: Given the sentence toward the end of the change:

> Spawns the worker only if the user requests it.

it might be worth directly stating that goroutines should not be started implicitly, but rather as a result of function call that could reasonably be expected to launch goroutine(s). (Ideally we could refer folks to [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), but that's a broader topic than this change.)

mway (commented):
This is great, thanks @abhinav!

style.md
alxn: I mean they are *guaranteed* to consume memory, everything else is a "maybe"/likely.

abhinav: Adapted what @mway said into a something extra explaining some of the concerns.

style.md
abhinav: ack. switched to,

```
Given a goroutine with a predictable lifetime,
or a goroutine that can be stopped with a signal,
there must be a way to wait for the goroutine to exit.
```

style.md
abhinav: My original intent was that this example only talks about how to build the ability to stop a goroutine into it,
and the next section talks about waiting for the goroutine to exit with either a WaitGroup or a channel. 

Let me see what else I can adjust here to make that clearer.

abhinav: Okay, seeing as I already say "there must be a predictable lifetime" and "in both cases there must be a way to wait for exit", I'll incorporate waiting for the exit in here similar to @mway's suggestion but with a channel instead of a wait group.

in the other sample where i talk about waiting for goroutines, i'll present both options.

abhinav:
Thanks for the feedback, all. Besides the minor comments, I've made the following changes:

- the original code sample is fully "correct" in that it includes both, how to stop a worker and how to wait for it to exit. 
- that you should always wait for the goroutine to exit is now a separate subsection demonstrating use of wait groups and channels for the purpose
- the "don't spawn goroutines in `init()`" section code sample now includes implementation of how you'd signal stop and wait for exit as well.

style.md
abhinav: I thought about this, but I'm not sure where to incorporate this.
It could be argued that if the returned object has a Close/Stop/Shutdown method, the goroutine is an implementation detail.
The contract is just that if you don't call the method when you're done, you're leaking resources.

mway: Yeah, fair. We can address all of this later (if ever).

style.md
```diff
@@ -1894,6 +1897,200 @@ this contract. Specifying field names inside tags makes the contract explicit,
...
+    time.Sleep(delay)
+  }
+}()
+```
+
+</td><td>
+
+```go
+stop := make(chan struct{}) // tells the goroutine to stop
+done := make(chan struct{}) // tells us that the goroutine exited
```
mway, line 1943: Are these intentionally ungrouped?

mway (approved):
Good stuff!

style.md
abhinav: Oops, yeah. Good catch.

PR #160, opened 2022-10-22
Update .golangci.yml

faustind:
This PR fixes the link to the example golangci config file in .golangci.yml.
Previously at https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml, now it's accessible at https://github.com/golangci/golangci-lint/blob/master/.golangci.reference.yml


abhinav (approved):

abhinav:
Thanks!

PR #164, opened 2023-02-01
style: Lint with markdownlint

pan93412:
I just noticed that I should open an issue before making a PR. Really sorry about not noticing it, and I have opened #165 for tracking this PR!

---

markdownlint is a static analysis tool with a library of rules to enforce standards and consistency for Markdown files. [^1]

This commit removes redundant spaces, unused references, indent mistakes, etc. from this document according to the report from `markdownlint`.

Note that I disable `MD033` (no-inline-html) as it is necessary for side-by-side code samples.

[^1]: https://github.com/DavidAnson/markdownlint

### DIFF
--- a/style.md
+++ b/style.md
@@ -48,6 +48,8 @@ row before the </tbody></table> line.
 
 -->
 
+<!-- markdownlint-disable MD033 -->
+
 # Uber Go Style Guide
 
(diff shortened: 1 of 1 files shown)


sywhang (approved):
Thank you!

PR #168, opened 2023-02-28
style.md: Split into separate files

abhinav:
Splits style.md into separate files
and stitches them back together with [stitchmd].

  [stitchmd]: https://github.com/abhinav/stitchmd

This makes it easier to manage the many sections of the style guide.
The src files are independently browseable
but this also keeps the final style.md with its old headers
so that existing users who link to it aren't broken.

Differences in style.md:

- *Prefer Specifying Container Capacity* no longer has its subitems
  listed in the table of contents.
  This is because those are part of the same document;
  the header itself doesn't have much content.
  This is fixable if desirable by either removing the content under the
  header, or by moving the slice and map sections into separate files.
  I chose the current configuration because it seemed fine as-is.
- All links have been converted to inline `[..](https://...)` style
  in the generated Markdown,
  but will be left alone in the input.
  This doesn't change how GitHub interprets these in any way.
- Multi-line list items are aligned to the list item delimiter.
  Again, this doesn't change how GitHub interprets these.

https://github.com/uber-go/guide/pull/168/files#diff-7b1bca67f13234851df27465985cb475d4a9ad49dc48df87a620c19378ddf03c

For automation, this includes a Makefile with two targets:

- `make style.md` updates the style guide from src/.
- `make lint` verifies it's up to date.

It sets up a GitHub Workflow that'll run `make lint` in CI,
and dependabot to update the dependencies in tools/
and for the workflow, on a periodic basis.

Refs https://github.com/uber-go/guide/discussions/159

PR #169, opened 2023-03-03
style.md: Add warning that the file is generated

abhinav:
Uses the -preface flag to add a comment to the top
which indicates that the file is generated
and that edits should be made to the src/ directory.

Lets us bring back the markdownlint comment
that was added in #164 and lost in #168.

Thanks to @mway for the suggestion.


blampe (approved):

mway (approved):
Nice!

PR #170, opened 2023-03-03
interface-receiver: Clarify why value receiver can be called

### LINKED ISSUE #163: Misleading formulation in "Receivers and Interfaces"
In `guide/style.md`:

The word **"only"** in the following comment is misleading and should be removed:

```
// You can only call Read using a value
```

In my understanding, this would mean that calling `Read` using a pointer is not possible, which is the opposite of the actual point of this section. 

See the french translation, for example, which reformulated it correctly.

abhinav:
The guidance in this section was a bit confusing
because it didn't quite explain why we could only call Read, not Write,
on the values stored in the map.

Resolves #163

### DIFF
--- a/src/interface-receiver.md
+++ b/src/interface-receiver.md
@@ -20,17 +20,22 @@ func (s *S) Write(str string) {
   s.data = str
 }
 
+// We cannot get pointers to value stored in maps.
 sVals := map[int]S{1: {"A"}}
 
-// You can only call Read using a value
+// We can call Read on values stored in the map because Read
+// has a value receiver.
 sVals[1].Read()
 
-// This will not compile:
+// We cannot call Write on values stored in the map because Write
+// has a pointer receiver, and it's not possible to get a pointer
+// to a value stored in a map.
+//
 //  sVals[1].Write("test")
 
 sPtrs := map[int]*S{1: {"A"}}
 
-// You can call both Read and Write using a pointer
+// You can call both Read and Write if the map stores pointers.
 sPtrs[1].Read()
 sPtrs[1].Write("test")
 ```
--- a/src/interface-receiver.md
+++ b/src/interface-receiver.md
@@ -20,11 +20,13 @@ func (s *S) Write(str string) {
   s.data = str
 }
 
-// We cannot get pointers to value stored in maps.
+// We cannot get pointers to values stored in maps, because they are not
+// addressable values.
 sVals := map[int]S{1: {"A"}}
 
 // We can call Read on values stored in the map because Read
-// has a value receiver.
+// has a value receiver, which does not require the value to
+// be addressable.
 sVals[1].Read()
 
 // We cannot call Write on values stored in the map because Write
@@ -35,7 +37,8 @@ sVals[1].Read()
 
 sPtrs := map[int]*S{1: {"A"}}
 
-// You can call both Read and Write if the map stores pointers.
+// You can call both Read and Write if the map stores pointers,
+// because pointers are intrinsically addressable.
 sPtrs[1].Read()
 sPtrs[1].Write("test")
 ```
--- a/style.md
+++ b/style.md
@@ -216,17 +216,22 @@ func (s *S) Write(str string) {
   s.data = str
 }
 
+// We cannot get pointers to value stored in maps.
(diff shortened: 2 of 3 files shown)


src/interface-receiver.md
```diff
@@ -20,17 +20,22 @@ func (s *S) Write(str string) {
   s.data = str
 }
 
+// We cannot get pointers to value stored in maps.
```
mway, line 23:
```suggestion
// We cannot get pointers to value stored in maps, because they are not
// addressable values.
```

src/interface-receiver.md
```diff
@@ -20,17 +20,22 @@ func (s *S) Write(str string) {
...
 sVals := map[int]S{1: {"A"}}
 
-// You can only call Read using a value
+// We can call Read on values stored in the map because Read
+// has a value receiver.
```
mway, line 27:
```suggestion
// has a value receiver, which does not require the value to
// be addressable.
```

src/interface-receiver.md
```diff
@@ -20,17 +20,22 @@ func (s *S) Write(str string) {
...
+// We cannot call Write on values stored in the map because Write
+// has a pointer receiver, and it's not possible to get a pointer
+// to a value stored in a map.
+//
 //  sVals[1].Write("test")
 
 sPtrs := map[int]*S{1: {"A"}}
 
-// You can call both Read and Write using a pointer
+// You can call both Read and Write if the map stores pointers.
```
mway, line 38:
```suggestion
// You can call both Read and Write if the map stores pointers,
// because pointers are intrinsically addressable.
```

mway (commented):
Looks good overall, I think it might be worth referencing addressability by name because it's a useful concept (noting that you're basically describing it here in each scenario). What do you think?

style.md
```diff
@@ -216,17 +216,22 @@ func (s *S) Write(str string) {
   s.data = str
 }
 
+// We cannot get pointers to value stored in maps.
```
sywhang, line 219:
```suggestion
// We cannot get pointers to values stored in maps.
```

sywhang:
> I think it might be worth referencing addressability by name because it's a useful concept 

I'm +1 on this, it's a source of many common mistakes.

mway (approved):
Thank you!

src/interface-receiver.md
```diff
@@ -20,17 +20,25 @@ func (s *S) Write(str string) {
...
 //  sVals[1].Write("test")
 
 sPtrs := map[int]*S{1: {"A"}}
 
-// You can call both Read and Write using a pointer
+// You can call both Read and Write if the map stores pointers,
+// because pointers are intrinsically addressable.
 sPtrs[1].Read()
 sPtrs[1].Write("test")
 ```
```
peterbourgon, line 44: This may compile but it is not safe. `sPtrs[1]` will return a nil value if `1` is not in the map, and `s.Write` does not guard against a nil method receiver. (Nor should it: methods should be able to assume their receivers are valid, in general.)

Expressions of the form `m[key].Method()` which chain map value lookups with method calls on those values are almost never safe. It would be great if this example could be removed from the guide.

abhinav: Agreed that this is an unsafe form and `m[key].Method()` should generally be avoided.
However, I think this case was specifically demonstrating addressability of values -- the value stored in the map not being addressable, and therefore, not usable for the method.
I don't immediately have an alternative suggestion for how else to demonstrate addressable versus non-addressable right after the prior value-in-map example.

But again, I agree that we don't want to encourage `m[key].Method()`, so, short of a different way to make the point this is making, maybe a comment would suffice? Something like "This sample is just to demonstrate the non-addressable nature of values stored in maps. We don't recommend calling methods on map values directly. See [link]" -- possibly accompanied by a new style guide entry that discourages `m[key].Method()` -- although we'll probably want to discuss it with the maintainers in a new issue. I think there's some nuance to it besides "don't call methods on map items directly," e.g. "unless you're absolutely certain" or "unless you control all keys". Perhaps "be wary" is better than "don't", but I'll defer that to the discussion.

ttando (commented):
gh pr checkout 170

PR #171, opened 2023-03-05
Fix link to empty slices

cgrinds:


abhinav (approved):
Thanks!

abhinav:
@cgrinds You'll need to run `make style.md` to regenerate the file at the root of the repo.
(The CI check was broken; fixed in #172.)

Separately: the same broken link exists in src/var-decl.md if you'd like to include the fix for that in this change. Otherwise, I'm happy to follow up afterwards.

cgrinds:
@abhinav Thanks for the heads up on #172. Ran `make style.md` and updated PR.

The [link](https://github.com/uber-go/guide/blob/f94e5eac81a2938955393f872229d66e0bdf5875/src/var-decl.md?plain=1#L25-L27) in `src/var-decl.md` looks correct.

mway (approved):
Thank you!

PR #172, opened 2023-03-05
make: Fix lint target

abhinav:
The lint target currently fails silently in CI
because of the use of `[[`.

    /bin/sh: 2: [[: not found

This is because `[[` is a built-in in bash, not in sh.
sh only has `[`.

Fix this by setting the shell to bash in make.


mway (approved):
Thanks!

abhinav:
Sample bad change: https://github.com/uber-go/guide/actions/runs/4336938533/jobs/7572595150#step:4:1

PR #174, opened 2023-03-15
fix(table-test): make guidance less presciptive

### LINKED ISSUE #173: proposal: table test guidance is too prescriptive
Table tests are great for reducing repetition, but as the SUT becomes more complex, they become harder to read, maintain, and debug.

Making tests harder to read and maintain at scale negatively impacts velocity, productivity, and overall sentiment on testing.

One common pattern I notice is with logical branches based on variables like `shouldError`, `shouldCallExecutor`, `shouldDoXYZ`. This creates an extremely difficult-to-read set of tests. As an exmple, consider the following test:

```
tests := []struct{
  give     string
  wantHost string
  wantPort string
  wantErr error
  shouldCallXYZ bool
  giveXYZResponse string
}{
  // ...
}

for _, tt := range tests {
  t.Run( ... ) {
    err := DoThing(tt.stuff)
    if tt.wantErr != nil {
      require.EqualError(t, err, tt.wantErr)
    }
    require.NoError(t, err)
    if tt.shouldCallXYZ{
      xyzMock.EXPECT().Call(...).Return(tt.giveXYZResponse, nil)
    }
  })
}
```
These tests are confusing. It's hard to tell what's going to be checked in which conditions, and unclear when certain fields of the `tests` struct are even used. The reader has to reference logic within the `tests` structs, as well as the complex behavioral logic within the test loop to understand what's happening.

Consider again that I need to add another subsystem to the service that has a mock which is called only some of the time. This adds another layer of complexity to the table test, and potentially obstructs all previously written parts of the table.

Having the guidance push to use Table Tests for "repeated patterns" actually can negatively impact code quality. Rather than recommending Table Tests for such patterns, we should recommend using Table tests for situations when **ONLY the inputs and outputs of the system differ**. It would also be good to recommend to **AVOID** Table Tests with logic existing inside the test loop.

Aside:
Somewhat unrelated to style, but it's also harder to maintain tests at scale with the dynamic assignment of subtest names. Splitting up table tests more granularly into separate `TestXXX` functions makes it easier to keep track of the speed and reliability of tests.

tyler-french:
Resolves #173 

This PR adjusts the recommendations in the Table Test section to be less prescriptive and promotes usage that better aligns with readability and maintainability.


moisesvega:
Let's relax the example inputs and create a small table just like our other bad/good tables. 

Bad:
```
// func TestOpenFile (t *testingT)
tests := []struct{
  give      string
  want     string
  wantErr error
  // ...
}{
  // ...
}

for _, tt := range tests {
 ... 
  got, err := os.OpenFile(give)
  if tt.wantError != nil {
    require.Error(t, err)
    assert.ErrorIs(t, err, os.ErrNotExist)
    continue
  }
  require.NoError(t, err)
  assert.Equal(t, want, f.Name())
}
```

Good: 
```
// func TestOpenFileSuccess (t *testingT)
tests := []struct{
  give      string
  want     string
  // ...
}{
  // ...
}
for _, tt := range tests {
 ... 
  got, err := os.OpenFile(give)
  require.NoError(t, err)
  assert.Equal(t, want, f.Name())
  ...
}

// func TestOpenFileError (t *testingT)
tests := []struct{
  give      string
  want     err
  // ...
}{
  // ...
}
for _, tt := range tests {
 ... 
  _, err := os.Open(give)
  require.Error(t, err)
  assert.ErrorIs(t, err, os.ErrNotExist)
  ...
}
```

moisesvega:
Here is a better example of this behavior:

```
// func TestCall(t *testing.T) {
 errNotFound := errors.New("not found")
 tests := []struct {
     give       string
     want      string
     wantErr error
     setupMock func(m *mocks)
 }{
     {
         desc: "success",
         setupMocks: func(m *mocks) {
             m.EXPECT().Call().Return("success", nil)
             ...
         },
     },
     {
         desc: "error/mock",
         want: errNotFound,
         setupMocks: func(m *mock) {
             m.EXPECT().Call().Return("", errNotFound)
             ...
         },
     },
 }


 for _, tt := range tests {
     ...
     m := tt.setupMock(m)
     got, err := m.Call()
     if tt.wantErr != nil {
       require.Error(t, err)
       assert.ErrorIs(t, err, tt.wantErr)
       continue
     }
     require.NoError(t, err)
    assert.Equal(t, tt.want, got) 
 }

``

File: src/test-table.md [hunk 1]
```diff
@@ -1,7 +1,11 @@
 # Test Tables
 
-Use table-driven tests with [subtests] to avoid duplicating code when the core
-test logic is repetitive.
+Table-driven tests with [subtests] can be a helpful pattern for writing tests
+to avoid duplicating code when the core test logic is repetitive.
+
+If a system under test needs to be tested against _multiple conditions_ where
+*only* the **inputs** and **outputs** change, a table-driven test *can* and **should**
+be used to reduce redundancy and improve readability.
```
prashantv, line 8: couple of nits:
 * I think we can tone it down a little (e.g., there's a lot of emphasis *only*, *can*, *should*)
 * I think it's useful to provide some context for why

prashantv (commented):
Fwiw, I disagree with this recommendation.

I find it useful to have error and non-error cases in the same test table, so when looking at cases, you can see the small changes that cause something to become invalid/valid.

I think having simple test-logic is important to ensure the test is correct, and don't think that `wantErr` is complex enough to break correctness -- it's also a commonly used [pattern in the stdlib](https://cs.github.com/?scopeName=All+repos&scope=&q=repo%3Agolang%2Fgo+wantErr).

I don't think the "Bad" example is bad in this case, it's pretty reflective of real usage of test tables.

That said, the test runner should be simple: avoiding nested ifs, many ifs. Basically, the test runner shouldn't be so complex that it's not trivial to follow.

abhinav:
> Fwiw, I disagree with this recommendation.
> 
> I find it useful to have error and non-error cases in the same test table, so when looking at cases, you can see the small changes that cause something to become invalid/valid.

I agree that just forbidding sharing of tables for success and failure cases is bad.
But I think it might be possible to express that nuance here better while pushing the policy of clean tables.

tyler-french:
Converting to draft to address feedback

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+Large complex table tests harm readability and maintainability because test readers will
+have difficulty debugging test failures that occur.
+
+Table tests like this should be split into either multiple test tables or multiple 
+individual `Test...` functions
+
+Some general guidelines to follow are:
+
+* Tests should focus on the narrowest unit of behavior possible
+* Tests should contain no `if` statements/logic if at all possible
```
abhinav, line 122: that might be too strict but i agree with the point.
how about
```suggestion
* Tests should aim to minimize `if` statements and other control flow if at all possible
```

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+logic forks within the subtest `for` loop. Table tests should **NOT** be used whenever 
+there needs to be complex logic added inside this execution loop.
+
+Large complex table tests harm readability and maintainability because test readers will
+have difficulty debugging test failures that occur.
+
+Table tests like this should be split into either multiple test tables or multiple 
+individual `Test...` functions
+
+Some general guidelines to follow are:
```
abhinav, line 119: Suggestion:
```suggestion
Some ideals to aim for are:
```

That makes it clear that the "perfect test" will meet the following conditions but there's leeway as the authors deem fit.

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+
+* Tests should focus on the narrowest unit of behavior possible
+* Tests should contain no `if` statements/logic if at all possible
+* All table fields should be used in all table tests
+* All test logic should be evaluated for all table cases
```
abhinav, line 124: suggestion: make this more instructional and in an active voice:

```suggestion
* Focus on the narrowest unit of behavior
* Minimize `if` statements and other control flow if possible
* Ensure that all table fields are used in all tests
* Ensure that all test logic runs for all table cases
```

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+
+Simple logic, such as `shouldErr` and `expectErr`, that does not significantly impact readability,
+is acceptable as long as the tests are clear and understandable.
+
+A common clear pattern to avoid is the addition of `if` statements for fields such as mock expectations
+and expected behavior that test too much responsibility for a single test.
```
abhinav, line 130: We can connect this to the previous list of Do's if we put the "avoid" case first:

```suggestion
Concretely, that means that you should avoid
having multiple branching pathways (e.g. `shouldError`, `expectCall`, etc.),
addition of `if` statements for specific mock expectations (e.g. `shouldCallFoo`),
or placing functions inside the table (e.g. `setupMocks func(*FooMock)`).

As an exception, if the test body is short and straightforward,
it's acceptable to have a single branching pathway for success versus failure cases
with a table field like `wantErr` to specify error expectations.
```

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+		giveYErr      error
+	}{
+		// ...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.give, func(t *testing.T) {
+			err := DoThing(tt.stuff)
+			if tt.wantErr != nil {
+				require.EqualError(t, err, tt.wantErr)
```
abhinav, line 155: should return after this

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+			}
+			require.NoError(t, err)
+			if tt.shouldCallX {
+				xyzMock.EXPECT().Call().Return(tt.giveXResponse, tt.giveXErr)
+			}
+			if tt.shouldCallY {
+				xyzMock.EXPECT().Call().Return(tt.giveYResponse, tt.giveYErr)
```
abhinav, line 162: mocks should be set before call

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+			}
+		})
+	}
+}
+```
+This makes the test more difficult to change, to understanding, and to review correctness.
```
abhinav, line 168: typo:

```suggestion
This makes the test more difficult to change, understand, and review correctness of.
```

src/test-table.md
```diff
@@ -100,6 +104,75 @@ for _, tt := range tests {
...
+
+There are no strict guidelines, but readability and maintainability should always be
+top-of-mind when deciding on use of Table Tests vs. separate tests for multiple inputs/outputs
```
abhinav, line 171: I'm not sure whether it's `vs.` or `vs`, so I'll just suggest `versus` here:

```suggestion
top-of-mind when deciding on use of Table Tests versus separate tests for multiple inputs/outputs 
```

abhinav (commented):
Very nice! Thanks for the update @tyler-french.
I realize that this is still a draft, but I gave it a quick pass before we post it for others to review.

tyler-french:
> Very nice! Thanks for the update @tyler-french. I realize that this is still a draft, but I gave it a quick pass before we post it for others to review.

Thanks for the review! I left it as draft so that I could try to incorporate more feedback from @mway

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
 }
 ```
 
+### Avoid Messy Table Tests
+
+Table tests become difficult to read and maintain if there becomes significant
+logic forks within the subtest `for` loop. Table tests should **NOT** be used whenever 
+there needs to be complex logic added inside this execution loop.
```
mway, line 111: nit:

- s/if there becomes significant/if there are significant/
- s/within the subtest `for`/within the subtest's `for`/

Alternatively:

```suggestion
Table tests can be difficult to read or maintain if their subtests contain
conditional assertions or other branching logic. Table tests should **NOT** be
used whenever there needs to be complex or conditional logic inside subtests.
```

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
...
+
+Large complex table tests harm readability and maintainability because test readers will
```
mway, line 113:
```suggestion
Large, complex table tests harm readability and maintainability because test readers may
```

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
...
+have difficulty debugging test failures that occur.
+
+Table tests like this should be split into either multiple test tables or multiple 
+individual `Test...` functions
```
mway, line 117:
```suggestion
individual `Test...` functions.
```

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
...
+Some ideals to aim for are:
+
+* Focus on the narrowest unit of behavior
+* Minimize `if` statements and other control flow (if possible)
+* Ensure that all table fields are used in all tests
+* Ensure that all test logic runs for all table cases
+
+Concretely, that means that you should avoid
+having multiple branching pathways (e.g. `shouldError`, `expectCall`, etc.),
+addition of `if` statements for specific mock expectations (e.g. `shouldCallFoo`),
```
mway, line 128:
```suggestion
using `if` statements for specific mock expectations (e.g. `shouldCallFoo`),
```

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
...
+				xMock.EXPECT().Call().Return(tt.giveXResponse, tt.giveXErr)
+			}
+			if tt.shouldCallY {
+				yMock.EXPECT().Call().Return(tt.giveYResponse, tt.giveYErr)
+			}
+		})
+	}
+}
+```
+This makes the test more difficult to change, understand, and review correctness of.
```
mway, line 180:
```suggestion
This complexity makes it more difficult to change, understand, and prove the correctness of the test.
```

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
...
+
+There are no strict guidelines, but readability and maintainability should always be
+top-of-mind when deciding on use of Table Tests versus separate tests for multiple inputs/outputs 
+to a system.
```
mway, line 184:
```suggestion
While there are no strict guidelines, readability and maintainability should always be
top-of-mind when deciding between Table Tests versus separate tests for multiple inputs/outputs to a system.
```

mway (commented):
This is looking great, thanks @tyler-french!

Left a few preemptive nits - feel free to use/amend what you think makes sense.

src/test-table.md
```diff
@@ -100,6 +104,87 @@ for _, tt := range tests {
 }
 ```
 
+### Avoid Messy Table Tests
```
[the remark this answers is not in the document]
tyler-french: I agree that "Complexity" would probably more clear/concrete from the perspective of a reader trying to understand when to use this.

src/test-table.md
```diff
@@ -100,6 +104,89 @@ for _, tt := range tests {
...
+				require.EqualError(t, err, tt.wantErr)
+				return
+			}
+			require.NoError(t, err)
+			if tt.shouldCallX {
+				xMock.EXPECT().Call().Return(tt.giveXResponse, tt.giveXErr)
+			}
+			if tt.shouldCallY {
+				yMock.EXPECT().Call().Return(tt.giveYResponse, tt.giveYErr)
+			}
```
abhinav, line 175: These EXPECT calls should also be in the "setup mocks" section.
gomock (which is what this looks like) wants expectations set in advance.
```suggestion
			yMock := xmock.NewMockY(ctrl)
			if tt.shouldCallX {
				xMock.EXPECT().Call().Return(tt.giveXResponse, tt.giveXErr)
			}
			if tt.shouldCallY {
				yMock.EXPECT().Call().Return(tt.giveYResponse, tt.giveYErr)
			}

			got, err := DoComplexThing(tt.give, xMock, yMock)

			// verify results
			if tt.wantErr != nil {
				require.ErrorIs(t, err, tt.wantErr)
				return
			}
			require.NoError(t, err)
```

abhinav (approved):
LGTM w minor comments.

@tyler-french thanks a lot for doing this.
Be sure to run `make style.md` to update the top-level style guide file.

This is good as-is, but if you'd prefer for this item to appear in the Table of Contents, you'll want to move it to its own file (test-table-simple.md?) but that's not necessary: we don't have the other subsection (parallel tests) in a separate file either. 

Please also wait for approval from @mway.

tyler-french:
> This is good as-is, but if you'd prefer for this item to appear in the Table of Contents, you'll want to move it to its own file (test-table-simple.md?) but that's not necessary: we don't have the other subsection (parallel tests) in a separate file either.

Yeah, I think it's ok to keep this under the "Table Tests" section.

abhinav (approved):

src/test-table.md
```diff
@@ -100,6 +104,88 @@ for _, tt := range tests {
...
+
+Some ideals to aim for are:
+
+* Focus on the narrowest unit of behavior
+* Minimize `if` statements and other control flow (if possible)
+* Ensure that all table fields are used in all tests
+* Ensure that all test logic runs for all table cases
+
+Concretely, that means that you should avoid
+having multiple branching pathways (e.g. `shouldError`, `expectCall`, etc.),
```
prashantv, line 127: nit: typically it's called `wantErr` (and the example below uses that too)

src/test-table.md
```diff
@@ -100,6 +104,88 @@ for _, tt := range tests {
...
+			if tt.wantErr != nil {
+				require.EqualError(t, err, tt.wantErr)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, want, got)
+		})
+	}
+}
+```
```
prashantv, line 179: Can we add the "Good" example of the above that allows testing the different cases too -- my worry is we're not seriously considering what the alternative would look like to test the same cases, and whether it's actually better.

prashantv (commented):
While I agree in principle with the guidance (keep test bodies simple), I worry that this guidance is pushing for separate test methods to avoid ifs in the test body, but doesn't take into account the impact to readability by splitting tests:
 * Test cases split across different test methods, making it hard to see differences between cases that are leading to different behaviour
 * Test runners will be duplicated, but likely some small differences, but in way that's harder to identify the differences.

A test with multiple mock methods is because the code under test is calling multiple methods -- have we tried going through an exercise of this guidance would affect tests for methods like that, and whether that's better for readability?

My personal preference is to group related test cases together to make it clear how a change to a test case affects behaviour. The test struct should make the runner logic obvious, if that's not the case, then it's probably too complex.

src/test-table.md
mway: I think the intent was different than `wantErr` (conventionally speaking) - `wantErr` is generally an `error` that you assert (e.g. with `errors.Is`), but `shouldError` is probably a boolean that implies branching for conditional assertion.

Having thought about it a bit more, I think we can probably remove this
paragraph and instead call out the things that are bad in the example below; that would help tie them to a concrete example ("how these things *can be* bad") without necessarily over-prescribing ("these things are *always* bad").

src/test-table.md
mway: +1, we should have both good/bad examples where possible, if only to illustrate contrast.

mway:
> While I agree in principle with the guidance (keep test bodies simple), I worry that this guidance is pushing for separate test methods to avoid ifs in the test body, but doesn't take into account the impact to readability by splitting tests:

I think it's OK to encourage decomposing tests into smaller units, but maybe we need a caveat? Something like:

> Be careful not to over-correct: for example, when testing behavior that
> changes based on input, it may be preferable to group similar cases
> together to better illustrate how behavior changes across all inputs,
> rather than splitting otherwise comparable units into separate tests
> and making them harder to compare and contrast.

This also brings to mind another subtlety that's been implied (sort of) but not directly discussed: test "depth", by which I mean "within a given test, the number of things successively asserted that require previous assertions to hold". Basically cyclomatic complexity, but specifically about relationships between assertions (which implies conditional assertion). Trivial example:

```go
func(t *testing.T) {
  // depth = 0; no dependency on previous assertions
  foo, err := NewFoo(tt.giveFooParam)
  require.ErrorIs(t, err, tt.wantFooError)

  if tt.wantFooError != nil {
    return
  }

  // depth = 1; depends on foo
  bar, err := foo.GetBar(tt.giveBarParam)
  require.ErrorIs(t, err, tt.wantBarError)

  if tt.wantBarError != nil {
    return
  }

  // depth = 2; depends on bar
  require.ErrorIs(t, bar.Baz(tt.giveBazParam), tt.wantBazError)
}
```

Through that lens, I think the goal is actually twofold: (1) minimizing depth, and by extension (2) avoiding conditional assertions. Conditions are probably fine if they're terminal, but once you start increasing depth (read: adding conditional assertions), that's when it gets hairy.

src/test-table.md
```diff
@@ -100,6 +104,88 @@ for _, tt := range tests {
...
+Large, complex table tests harm readability and maintainability because test readers may
+have difficulty debugging test failures that occur.
+
+Table tests like this should be split into either multiple test tables or multiple 
+individual `Test...` functions.
+
+Some ideals to aim for are:
+
+* Focus on the narrowest unit of behavior
+* Minimize `if` statements and other control flow (if possible)
```
mway, line 122: Based on [this comment](https://github.com/uber-go/guide/pull/174#issuecomment-1521829041):

```suggestion
* Minimize "test depth", and avoid conditional assertions (see below)
```

and then expand later, after the list:

> In this context, "test depth" means "within a given test, the number of
> successive assertions that require previous assertions to hold" (similar
> to cyclomatic complexity, except that it also describes prerequisite
> relationships between assertions).
>
> Having "shallower" tests means that there are fewer relationships between
> assertions and, more importantly, that those assertions are less likely
> to be conditional by default.

That might better clarify the actual goal (versus banning conditions in general). I think that this would also allow the guidance to be (1) applied more consistently and (2) more flexible around what a test should look like.

Thoughts?

src/test-table.md
tyler-french: Added a "good" example and tried to emphasize that this isn't a one-size-fits-all situation. The guidance should be more about avoiding complexity and confusion rather than any specific implementation in particular.

tyler-french:
@mway @prashantv @abhinav I tried to update the wording with some of @mway's suggestions, while also keeping some examples with the `if`.

I worry that getting too "theoretical" will hurt the experience for some readers. It's good to have a concrete sense of how something could look (e.g. multiple `if`s in the loop). Conversely, I do agree that saying "avoid `if`s" goes against the original intent of being less prescriptive.

Thanks again for the detailed comments and suggestions here, and thanks for the patience with the long turnaround 😄

mway (approved):
Looks great, thank you @tyler-french!

Once we sort CI this can be merged.

abhinav:
Do not merge yet. PR green is a false. Needs #183 first.

tyler-french:
Thanks for the reviews! And I don't have merge permission anyway here, so best of luck with the CI!

PR #179, opened 2023-04-10
Handle errors once

### LINKED ISSUE #65: Guideline: Don't Log and Return Errors
Creating this to track adding a general guideline around logging-vs-returning
errors, and doing only one of the two at any error handling site to avoid
double logging.

abhinav:
Creating as draft PR to get early feedback.

[Preview](https://github.com/abhinav/uber-go-guide/blob/error-once/src/error-once.md)

---

Adds new guidance on how handling errors once
rather than double handling, e.g. log-and-return.

Resolves #65

### DIFF
--- a/src/error-once.md
+++ b/src/error-once.md
@@ -1,0 +1,106 @@
+# Handle Errors Once
+
+When a function receives an error from a caller,
+it can handle it in a variety of different ways.
+These include, but not are limited to:
+
+- returning the error
+- [wrapping the error](error-wrap.md) and returning that
+- logging the error and returning nil
+- logging the error and returning a different error
+- logging the error and continuing running
+- matching the error with `errors.Is` or `errors.As`
+  and then handling the two error branches differently
+
+In general, a function should handle an error only once.
+It should not, for example, log the error and then return it again
+because its callers will likely handle the error too.
+
+As demonstrative examples, consider the following cases:
+
+<table>
+<thead><tr><th>Description</th><th>Code</th></tr></thead>
+<tbody>
+<tr><td>
+
+**Bad**: Log the error and return it
+
+Callers will likely handle the error as well--likely logging it before doing so
+causing a lot of noise in the application logs.
+
+</td><td>
+
+```go
+u, err := getUser(id)
+if err != nil {
+  // BAD: See description
+  log.Printf("Could not get user %q: %v", id, err)
+  return err
+}
+```
+
+</td></tr>
+<tr><td>
+
+**Good**: Wrap the error and return it
+
+Callers will be able to handle the error,
+matching with `errors.Is` or `errors.As` for specific errors.
+
+</td><td>
+
+```go
+u, err := getUser(id)
+if err != nil {
+  return fmt.Errorf("get user %q: %w", id, err)
+}
+```
+
+</td></tr>
+<tr><td>
+
+**Good**: Log the error and degrade gracefully
+
+If retrieving the value is not strictly necessary,
+we can provide a degraded but unbroken experience
+by using a fallback value.
+
+</td><td>
+
+```go
+u, err := getUser(id)
+if err != nil {
+  log.Printf("Could not get user %q: %v", id, err)
+  u = nil
+}
+
+```
+
+</td></tr>
+<tr><td>
+
+**Good**: Match the error and degrade gracefully
+
+If it's expected for retrieval to fail in specific cases,
+match on that error case and degrade gracefully.
+
+For all other cases, wrap the error and return it.
+Callers will be able to add their own special handling
+if necessary.
+
+</td><td>
+
+```go
+
+u, err := getUser(id)
+if err != nil {
+  if !errors.Is(err, ErrNotFound) {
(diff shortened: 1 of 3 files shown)


alxn:
The other note I like, by wrapping errors you are able to specifically test you are hitting that wrapped case. If you have a module that just passes through errors, future changes may cause those paths to change, and you to lose test coverage.

mway (commented):
Looks good so far. I'm split on whether it's worth discussing error boundaries (at a high level), maybe through examples (specifically thinking of your degradation examples).

The general idea being that some confusion around error handling may stem from not really understanding what should really "care" about an error (i.e., what should that error's terminal handler be), and examples that help folks think about how to model that (through the lens of error boundaries/SOC) may help as well.

Thoughts?

abhinav:
Could you clarify what you're imagining here, @mway?
I'm having trouble coming up with how to communicate the idea of error boundaries succinctly.
"Handle errors at abstraction boundaries" or variations of it seem a bit too high-level.

sywhang (commented):
Overall I like it. Though it may get a little blurry in terms of what errors are ok to handle and what aren't in the `Match the error and degrade gracefully` section without some kind of guidance on error boundaries as @mway suggested. 

For instance suppose someone has code similar to the example, except `getUser` is an external pkg:
```go
u, err := someClient.getUser(id)
```

Then does one match for canonical error types defined by `someClient` and nothing else, like:
```go
if errors.Is(err, someClient.ErrUserInactive) {
   // ...
}
``` 
, or does one check for any kind of commonly-occurring errors returned from `someClient.getUser` regardless of whether the error is wrapped or not
```go
if errors.Is(err, fs.PathError) {
   // ...
}
```

abhinav:
> does one match for canonical error types defined by someClient and nothing else, [..]
> or does one check for any kind of commonly-occurring errors returned from someClient.getUser regardless of whether the error is wrapped or not

Okay, yeah, this is where the confusion might be coming from.
I assumed it was obvious that you would check for whatever errors someClient.GetUser is documented as returning that you want to have special behaviors for. You will not check for `fs.FooError` if GetUser doesn't say that's a thing it can return, even if it uses that internally and possibly wraps it. 

I'll see if there's a way to expand the match errors section.

abhinav:
@mway @sywhang Updated to try to clarify error matching more.

I also want to clarify my intent here to help future discussion.
This is not intended to be a comprehensive list of different ways of handling errors.
It's a statement that when you handle errors (of which there are many ways, here are a few), you should do it only once.

If you think that that intent is getting lost because we enumerate some ways of handling errors, I'm happy to pare down the entry and drop those extraneous details.

src/error-once.md
```diff
@@ -0,0 +1,112 @@
+# Handle Errors Once
+
+When a function receives an error from a caller,
```
alxn, line 3: how does a function receive an error from a caller or am I being stupid here?

src/error-once.md
```diff
@@ -0,0 +1,112 @@
...
+  logging the error and degrading gracefully
+- if the error represents a domain-specific failure condition,
+  returning a well-defined error
+- [wrapping the error](error-wrap.md) and returning it
+- returning the error as-is
+
+Regardless of how the error is handled, generally,
+a function should handle an error only once.
+It should not, for example, log the error and then return it again
+because its callers will likely handle the error too.
```
alxn, line 22: oh, by handle you mean "take an action", where action 1 is log, and action 2 is return.

or you mean "a function should touch a received error only once."

src/error-once.md
```diff
@@ -0,0 +1,112 @@
+# Handle Errors Once
+
+When a function receives an error from a caller,
+it can handle it in a variety of different ways
+depending on what it knows about the error.
+
+These include, but not are limited to:
+
+- if the function contract defines specific errors,
```
prashantv, line 9: nit: function can refer to the caller or callee, what do you think of explicitly using those terms throughout? "if the callee function defines specific errors"

src/error-once.md
```diff
@@ -0,0 +1,112 @@
...
+  matching the error with `errors.Is` or `errors.As`
+  and handling the branches differently
+- if the error is recoverable,
+  logging the error and degrading gracefully
+- if the error represents a domain-specific failure condition,
+  returning a well-defined error
+- [wrapping the error](error-wrap.md) and returning it
+- returning the error as-is
```
prashantv, line 17: optional: since "wrapping the error" also talks about returning errors as-is so consider if these should be merged into one

src/error-once.md
abhinav: 🤦  oops. yeah, callee.

src/error-once.md
abhinav: I meant "take an action". Let me reword a tad bit.

src/error-once.md
abhinav: oh good call. will change.

mway (approved):
Nice, thanks @abhinav! LGTM modulo a couple quick suggestions.

PR #180, opened 2023-04-22
ci/stitchmd: Don't build from source, use action

abhinav:
This replaces the manual compilation and running of stitchmd
with the [stitchmd-action](https://github.com/abhinav/stitchmd-action).
The action downloads a pre-built copy of stitchmd from GitHub,
and runs it.

The action offers multiple execution modes. Those relevant here are:

- check: fail if the file is not up-to-date
- write: update the file if it's not up-to-date

The workflow added here uses both:

- For a push to any branch, it will run in check mode. Read-only.
  It'll check that the file is up-to-date and that's it.
- However, if a pull request is created with a branch,
  it will run in write mode -- automatically updating the PR if needed.

This has some nice properties in terms of UX and safety:

- We never push to a branch without permission.
  This alleviates safety concerns around automated pushes.
- We make it easy to make edits on the GitHub UI,
  and still have them be picked up by the system
  without creating a local checkout
- Setting up Go, building stitchmd from source, and then running it--all
  this takes some time.
  It's faster to download and run the binary.

The instructions taken to set up this dual mode
can be found here:
https://github.com/abhinav/stitchmd-action#automatically-update-for-prs-only

Note that this nukes the tools/ directory, but keeps the `Makefile`
so that if someone wants to run this locally, they still can.


mway (approved):
This is great, thank you!

PR #183, opened 2023-05-09
ci/stitchmd: Run git-auto-commit-action on pull_request_target

abhinav:
Auto-commit to PRs only if the event is pull_request_target not pull_request, because we've changed the event we dispatch on.

(I missed this in the prior two PRs. 😞 )


mway (approved):
third time's a charm, happy monday

PR #184, opened 2023-05-09
visual fix: reduce width and replace tabs

tyler-french:
The tables added in #174 are a bit wide and hard to fit on smaller screens.

By changing the `\t` to `  `, and splitting a line, we can reduce this width.

### DIFF
--- a/src/test-table.md
+++ b/src/test-table.md
@@ -146,9 +146,8 @@ it's acceptable to have a single branching pathway for success versus failure ca
 with a table field like `shouldErr` to specify error expectations.
 
 <table>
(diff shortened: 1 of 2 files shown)


tyler-french:
FYI @abhinav, I noticed it looks a little wide on some browsers: https://github.com/uber-go/guide/blob/master/style.md#test-tables

abhinav:
I tried transposing the table to see if that helped.
It takes less horizontal space, but I don't think that's better. I'll revert my change.

abhinav:
Yeah, I think tabs to two spaces here is the only way right now to demonstrate them side-by-side without a lot of horizontal scrolling.
We already use two spaces for indentation in most other code samples in the guide.

PR #186, opened 2023-06-05
Add advice on formatting strings using `%q`

### LINKED ISSUE #185: Proposal: prefer `%q` to format strings instead of `%s`
> **Note** All credit goes to @abhinav 

I learned this from one of @abhinav's amazing talks, and thought we could probably add this practice to this doc.

Whenever formatting messages that contain a string component via `fmt`, prefer `%q` instead of `%s`. This will wrap the specified string in quotes, helping it stand out from the rest of the error message. More importantly, if the string is empty, it will provide a more helpful error message.

<table>
<thead><tr><th>Bad</th><th>Good</th></tr></thead>
<tbody>
<tr><td>

```go
fmt.Errrof("file %s not found", filename)
// Prints the following: 
// open does_not_exist: invalid path
//
// Or if the string is empty:
// open : invalid path
```

</td><td>

```go
fmt.Errrof("file %q not found", filename)
// Prints the following:
// open "does_not_exist": invalid path
//
// Or if the string is empty:
// open "": invalid path
```

</td></tr>
</tbody></table>

yuxincs:
> **Note** All credit goes to @abhinav 

This PR adds suggestions on formatting strings using `%q` over `%s`.

Source: https://abhinavg.net/2021/12/29/fmt-errof-q/ by @abhinav 

Fixes #185

### DIFF
--- a/src/formatting-strings.md
+++ b/src/formatting-strings.md
@@ -1,0 +1,38 @@
+# Format Strings Using `%q`
+
+Whenever formatting messages that contain a string component via `fmt`, use `%q` instead of `%s`. This will wrap the specified string in quotes, helping it stand out from the rest of the error message. More importantly, if the string is empty, it will provide a more helpful error message.
+
+<table>
+<thead><tr><th>Bad</th><th>Good</th></tr></thead>
+<tbody>
+<tr><td>
+
+```go
+fmt.Errrof("file %s not found", filename)
+// Prints the following:
+// file myfile.go not found
+//
+// Or if the string is empty:
+// file not found
+```
+
+</td><td>
+
+```go
+fmt.Errrof("file %q not found", filename)
+// Prints the following:
+// file "myfile.go" not found
+//
+// Or if the string is empty:
+// file "" not found
+```
+
+</td></tr>
+</tbody></table>
+
+This advice applies more generally to other contexts when reporting user-specified data, such as logging invalid usernames:
+
+```go
+log.Printf("User %q does not exist", username)
+// User "no_name" does not exist
+```
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -60,6 +60,7 @@
   - [Initializing Maps](map-init.md)
   - [Format Strings outside Printf](printf-const.md)
   - [Naming Printf-style Functions](printf-name.md)
+  - [Format Strings Using `%q`](formatting-strings.md)
 - Patterns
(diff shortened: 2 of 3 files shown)


src/formatting-strings.md
```diff
@@ -0,0 +1,38 @@
...
+
+</td></tr>
+</tbody></table>
+
+This advice applies more generally to other contexts when reporting user-specified data, such as logging invalid usernames:
+
+```go
+log.Printf("User %q does not exist", username)
+// User "no_name" does not exist
+```
```
mway, line 38: I would generalize this slightly to something like:

> This advice can also apply more generally to other formatting situations
> when a dynamic string could confuse or change the formatting or meaning of
> surrounding text when interpolated.

Mostly to avoid bucketing "logging" as "formatting" (i.e., structured logging - which is at least as relevant as `log.Printf` scenarios - doesn't require the same treatment for fields, because they are already discrete values).

I think the `log.Printf` example is fine to keep, I would just put a good/bad example under the `fmt.Errorf` examples.

src/formatting-strings.md
```diff
@@ -0,0 +1,38 @@
+# Format Strings Using `%q`
+
+Whenever formatting messages that contain a string component via `fmt`, use `%q` instead of `%s`. This will wrap the specified string in quotes, helping it stand out from the rest of the error message. More importantly, if the string is empty, it will provide a more helpful error message.
```
mway, line 3: > Whenever formatting messages that contain a string component via `fmt`, use `%q` instead of `%s`.

I think we might want to clarify slightly: it's not necessarily just when a string component in general is used, but more when dynamic strings (incl. user-provided values) could potentially cause interpolation to result in something weird.

Maybe something like:
> Whenever dynamic strings - such as user-provided data - are used for formatting,
> prefer the placeholder `%q` over `%s`.

mway, line 3: > [...] helping it stand out from the rest of the error message.

> More importantly, if the string is empty, it will provide a more helpful error message.

I'd either generalize these slightly or remove the word "error" (this guidance can apply to more than just errors). Maybe something like:

> More importantly, if the string is empty or has its own formatting, the result
> will be clearer and more consistent.

mway (commented):
Thanks for the PR! Added a few suggestions, but looks good in general.

Ayokanmi-Adejola (approved):

PR #190, opened 2023-07-26
exit-once: Clarify that `run()` ins't prescriptive

abhinav:
Clarify that the example of using `run()` and `log.Fatal` isn't intended
to be prescriptive.

Provide samples of using `os.Exit` directly, using different exit codes,
and clarify that there's flexibility here on how everything is set up.

Refs #189


mway (approved):
Nice, LGTM. Thanks @abhinav!

PR #193, opened 2023-09-18
Update syntax in style.md

lcpulzone:
An apostrophe was missing from one word and placed in the wrong spot for another word. This PR adds and moves an apostrophe.

### DIFF
--- a/style.md
+++ b/style.md
@@ -79,7 +79,7 @@ misnomer, since these conventions cover far more than just source file
 formatting—gofmt handles that for us.
 
 The goal of this guide is to manage this complexity by describing in detail the
-Dos and Don'ts of writing Go code at Uber. These rules exist to keep the code
+Do's and Dont's of writing Go code at Uber. These rules exist to keep the code
 base manageable while still allowing engineers to use Go language features
 productively.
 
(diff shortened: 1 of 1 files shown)


mway:
Hey @lcpulzone, thanks for the PR!

The current state of the guide ("Dos and Don'ts", though "Do's and Don'ts" also appears to be acceptable) is aligned with guidelines and common use across different style guides and publications, as far as I can tell. Some examples:

- https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Plurals/faq0008.html
- https://dictionary.cambridge.org/dictionary/english/dos-and-don-ts
- https://www.economist.com/whichmba/application-dos-and-donts
- https://www.nytimes.com/2009/05/27/dining/27cbox.html
- https://developer.apple.com/design/tips/
- https://www.theguardian.com/law/2017/nov/21/the-dos-and-donts-of-writing-a-personal-statement-for-law

I think this is OK to leave as-is. Even if there are counter arguments, the intent is clear; it's aligned *enough* that I don't think we're blatantly misspelling it.

Also, note that `style.md` should not be updated directly - it is generated, and the source for this particular text is in [src/intro.md](https://github.com/uber-go/guide/blob/master/src/intro.md).

PR #210, opened 2024-02-20
Deleted duplicated line in style.md

mihailstefanov:
deleted duplicated line "- add context with `fmt.Errorf` and the `%v` verb"


mihailstefanov:
My mistake was %v and %w, probably I need glasses.

Issue #216, opened 2024-08-01
Add more guidance on scope for constants


tyler-french:
Hoping to get some opinions on this.

One pattern I see a lot is with making **global** constants for values that are only used within a certain function.

For example

```
const _defaultPort = 8000

func something() {
  doSomething(_defaultPort)
  port := _defaultPort
}
```

In my opinion, there should be an adjustment to the **# Reduce Scope of Variables** section to include more emphasis on what to do with **Constants**.

I mostly want to get opinions here, but I generally would prefer to change the above code to the following:
``````
func something() {
  const defaultPort = 8000
  doSomething(defaultPort)
  port := defaultPort
}
```
Thoughts?

tyler-french:
Maybe a follow-up, what about things like `types` and enums:
```
func doSomething(input string) {
	type x string
	const (
		y x = "1"
		z x = "1"
	)
	...
```

sywhang:
https://github.com/uber-go/guide/blob/master/style.md#reduce-scope-of-variables kind of covers this, IMO. Perhaps adding some more examples with types/consts can be helpful?

PR #217, opened 2024-08-01
give guidance to reduce scope of constants

tyler-french:
See https://github.com/uber-go/guide/issues/216 for more information.

This is just a draft, and more to start discussion to understand what the correct guidance here is.

### DIFF
--- a/src/var-scope.md
+++ b/src/var-scope.md
@@ -67,7 +67,8 @@ return nil
 </td></tr>
 </tbody></table>
 
-Constants do not need to be global unless they are used in multiple functions or files.
+Constants do not need to be global unless they are used in multiple functions or files
+or are part of an external contract of the package.
 
 <table>
 <thead><tr><th>Bad</th><th>Good</th></tr></thead>
--- a/src/var-scope.md
+++ b/src/var-scope.md
@@ -93,7 +93,7 @@ func Bar() {
     defaultPort = 8080
     defaultUser = "user"
   )
-  fmt.Println("Default port", _defaultPort)
+  fmt.Println("Default port", defaultPort)
 }
 ```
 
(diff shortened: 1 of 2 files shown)


src/var-scope.md
```diff
@@ -1,6 +1,6 @@
-# Reduce Scope of Variables
+# Reduce Scope of Variables and Constants
```
tyler-french, line 1: This might mess up existing links?

If so, we can just update description and examples

sywhang: Yeah let's not edit the heading if possible. I think the description is pretty clear.

src/var-scope.md
```diff
@@ -66,3 +66,36 @@ return nil
 
 </td></tr>
 </tbody></table>
+
+Avoid making constants global unless they are used in multiple functions.
```
sywhang, line 70:
```suggestion
Constants do not need to be global unless they are used in multiple functions or files.
```

src/var-scope.md
```diff
@@ -66,3 +66,36 @@ return nil
...
+
+</td><td>
+
+```go
+func Bar() {
+  const (
+    defaultPort = 8080
+    defaultUser = "user"
+  )
+  fmt.Println("Default port", _defaultPort)
```
abhinav, line 96:
```suggestion
  fmt.Println("Default port", defaultPort)
```

src/var-scope.md
```diff
@@ -66,3 +66,36 @@ return nil
 
 </td></tr>
 </tbody></table>
+
+Constants do not need to be global unless they are used in multiple functions or files.
```
abhinav, line 70: optional qualification:

```suggestion
Constants do not need to be global unless they are used in multiple functions or files
or are part of an external contract of the package.
```

abhinav (approved):

PR #232, opened 2025-08-19
Fix reference to variable scoping inside parallel t.Run executions

gonzaloserrano:
Since Go 1.22 that's not needed anymore, see https://tip.golang.org/doc/go1.22#language


abhinav (commented):
I think we can just delete the parallel section? WDYT @sywhang @prashantv

prashantv:
> I think we can just delete the parallel section? WDYT @sywhang @prashantv

Agree

gonzaloserrano:
done

abhinav (approved):

PR #251, opened 2025-12-22
keep example the same

jjpinto:


albertyw (changes requested):
The changes should be made in the `src` directory.  The `style.md` file is autogenerated.

PR #257, opened 2026-02-13
Fix pointer typo

Leech1:


albertyw (approved):
Need to undo the second auto-fix commit

albertyw (changes requested):
The changes should be made in the `src` directory.  The `style.md` file is autogenerated.

